PR #6, opened 2023-09-15
Vertex arrays for BMD::RenderMesh

sven-n:
It was hard to find any examples on this on the internet how to use vertex arrays in legacy open gl versions. Found this one for OpenTK and adapted it back to C++:
https://gdbooks.gitbooks.io/legacyopengl/content/Chapter8/va_p5.html

Basically, instead of calling a lot of opengl functions for each vertex, we call one with all the required data in some arrays.
Further improvements require to use VAO and VBO, or newer methods (shaders). However, these would require a major refactor/rewrite of the game clients rendering code.


sven-n:
Didn't do any benchmarks yet, but it's working and shouldn't be slower by common sense.

So, I'll merge it for now.

PR #7, opened 2023-09-24
Refactor how packets are sent + Unicode support

sven-n:
The goal is to get rid of wsclientinline.h in the future to improve readability and maintainability of the code.

For this, I made the extension methods of MUnique.OpenMU.Network also available on the C++ side, so that the code can use these instead of including macros of wsclientinline.h.
There is a lot of interop glue code required, which is auto-generated by some XSLT, based on the XML packet definitions data of OpenMU.

After this PR, the other side of the connection will be refactored - receiving packets. This means, that when it's done, the XML is the single source of truth for all network messages which are exchanged between client and server. Then we can be sure that they always fit together by 100 %.

Anyway, one usage example:
Old way: `SendRequestServerList();` included a macro from wsclientinline.h
New way: `SocketClient->ToConnectServer()->SendServerListRequest();`: ToConnectServer() returns a pointer to an object which offers all the possible messages which are available on the connect server. Calling SendServerListRequest for example, will call the SendServerListRequest on the ManagedLibrary. Instead of ToConnectServer(), there is also ToChatServer() and ToGameServer().
grafik

As a side product, I changed the client to use unicode strings in memory (wchar_t and std:wstring to keep interop simple), treating strings in packets and files as UTF-8. This has several upsides, since we can now support all kind of languages without messing around with ANSI codepages. And it can help implement new features like emojis, too. For now, some emojis can be displayed already, but only in 2 colors.


sven-n:
It's getting closer to completion, so I set it to be 'ready for review'.

Source Main 5.2/source/ZzzInterface.cpp
```diff
@@ -592,39 +601,46 @@ void RenderNotices()
     NoticeInverse += FPS_ANIMATION_FACTOR;
 }
 
-void CutText(const char* Text, char* Text1, char* Text2, int Length)
+void CutText(const wchar_t* Text, wchar_t* Text1, wchar_t* Text2, int Length)
```
sven-n, line 604: This method needs to be fixed... it causes some empty chat lines, with just " : "

PR #16, opened 2024-04-20
Fix Lorencia Effect Main

Odisey25:
removes the effect of looking like snow or spring in lorencia

### DIFF
--- a/Source Main 5.2/source/MapManager.cpp
+++ b/Source Main 5.2/source/MapManager.cpp
@@ -1555,7 +1555,7 @@ void CMapManager::LoadWorld(int Map)
 
         swprintf(FileName, L"%s\\leaf01.tga", WorldName);
         LoadBitmap(FileName, BITMAP_LEAF1, GL_NEAREST, GL_CLAMP_TO_EDGE, false);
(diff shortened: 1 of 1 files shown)


Source Main 5.2/source/MapManager.cpp
```diff
@@ -1553,9 +1553,10 @@ void CMapManager::LoadWorld(int Map)
         swprintf(FileName, L"%s\\TileGrass03.tga", WorldName);
         LoadBitmap(FileName, BITMAP_MAPGRASS + 2, GL_NEAREST, GL_REPEAT, false);
 
-        swprintf(FileName, L"%s\\leaf01.tga", WorldName);
+        
+        swprintf(FileName, L"%s\\leaf01.tga", WorldName); 
         LoadBitmap(FileName, BITMAP_LEAF1, GL_NEAREST, GL_CLAMP_TO_EDGE, false);
-        swprintf(FileName, L"%s\\leaf01.jpg", WorldName);
+        swprintf(FileName,(Map==0||Map==3||Map==63)? L"%s\\leaf01.tga": L"%s\\leaf01.jpg",WorldName);
```
sven-n, line 1559: that makes sense - jpeg has no alpha channel - good catch :)

sven-n:
Thank you, that's something I wanted to fix for a long time xD

PR #17, opened 2024-04-21
Slider LayerDep Position

Odisey25:
the slider is at the top
qeWoGzr


sven-n:
A question: Isn't this sliding message supposed to be behind the UI controls (map name, mu helper buttons)? Right of them is enough space to read it, isn't it?

PR #20, opened 2024-04-21
Up Chatlog

Odisey25:
cicyGIg

change position of chatlog to the top


sven-n:
what's the benefit of that? I separated chat and system messages - are they still separate when you expand the chat box?

Odisey25:
> what's the benefit of that? I separated chat and system messages - are they still separate when you expand the chat box?
all messages go to the top 
example
MuServer 97d+99i - Ajuda em MuOnline (1)

sven-n:
Sorry, it's not compatible with my source:

grafik

PR #26, opened 2024-09-21
Small bugs & improvements

recone83:
Monster name length, MuMain was set to 32, but Land_End had names longer than 32 - caused crashes
Enabled Karutan

### DIFF
--- a/Source Main 5.2/source/MapManager.cpp
+++ b/Source Main 5.2/source/MapManager.cpp
@@ -14,6 +14,11 @@
 #include "GM3rdChangeUp.h"
 #include "GMCryWolf1st.h"
 #include "GMDoppelGanger2.h"
+
+#ifdef ASG_ADD_MAP_KARUTAN
+#include "GMKarutan1.h"
+#endif
+
 #include "LoadData.h"
 #include "NewUISystem.h"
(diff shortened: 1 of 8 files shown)


Source Main 5.2/source/MapManager.cpp
```diff
@@ -1199,10 +1204,10 @@ void CMapManager::Load() // OK
 #ifdef ASG_ADD_MAP_KARUTAN
         else if (IsKarutanMap())
         {
-            Models[66].Actions[0].PlaySpeed = 0.15f;
-            Models[66].Actions[1].PlaySpeed = 0.15f;
+            //Models[66].Actions[0].PlaySpeed = 0.15f;
```
sven-n, line 1207: why these changes?

sven-n:
I merged my other PR #25 which also enabled karutan 😅 I'll change monster name to length of 40, too.

Source Main 5.2/source/MapManager.cpp
recone83: Models does not exists, causes crash

recone83:
> I merged my other PR #25 which also enabled karutan 😅 I'll change monster name to length of 40, too.

sweet, thanks

Issue #12, opened 2023-10-15
Too many requests for the Pet Info


sven-n:
When moving the mouse over a pet (Dark Horse, Dark Raven) in the inventory, the client requests the current pet information (level, etc.). Unfortunately, this request is sent repeatedly for each frame.

It would be beneficial if the amount of requests could be minimized to a reasonable level - ideally, once per hover operation, with a timeout of about a second for the next request.

sven-n:
Found another reason, why this should be fixed. With every pet information from the server, all of the character stats are recalculated. In the debugger, the game stucks.

nitoygo:
Could this perhaps be aggregated in the server side?

nitoygo:
Idea: send the request only after opening the inventory.

PR #34, opened 2024-10-19
Fixed some data type or size exceptions

nitoygo:
* Fixed exception due to data size during kills
* Fixed exception due to data size when receiving level up packet 
* Fixed exception due to data size when equipping raven

### DIFF
--- a/Source Main 5.2/source/NewUIMyInventory.cpp
+++ b/Source Main 5.2/source/NewUIMyInventory.cpp
@@ -328,8 +328,8 @@ bool CNewUIMyInventory::IsEquipable(int iIndex, ITEM* pItem)
 
     if (pItem->Type == ITEM_DARK_RAVEN_ITEM)
     {
-        const PET_INFO* pPetInfo = GetPetInfo(pItem);
-        const WORD wRequireCharisma = (185 + (pPetInfo->m_wLevel * 15));
+        auto pPetInfo = GetPetInfo(pItem);
+        auto wRequireCharisma = (185 + (pPetInfo->m_wLevel * 15));
         if (wRequireCharisma > wCharisma)
             return false;
     }
--- a/Source Main 5.2/source/NewUIMyInventory.cpp
+++ b/Source Main 5.2/source/NewUIMyInventory.cpp
@@ -328,10 +328,17 @@ bool CNewUIMyInventory::IsEquipable(int iIndex, ITEM* pItem)
 
     if (pItem->Type == ITEM_DARK_RAVEN_ITEM)
     {
-        auto pPetInfo = GetPetInfo(pItem);
-        auto wRequireCharisma = (185 + (pPetInfo->m_wLevel * 15));
-        if (wRequireCharisma > wCharisma)
+        const auto pPetInfo = GetPetInfo(pItem);
+        if (pPetInfo->m_dwPetType == PET_TYPE_NONE)
+        {
             return false;
+        }
+
+        const auto requiredCharisma = (185 + (pPetInfo->m_wLevel * 15));
+        if (requiredCharisma > wCharisma)
+        {
+            return false;
+        }
     }
 
     if (gMapManager.WorldActive == WD_7ATLANSE && (pItem->Type >= ITEM_HORN_OF_UNIRIA && pItem->Type <= ITEM_HORN_OF_DINORANT))
--- a/Source Main 5.2/source/WSclient.cpp
+++ b/Source Main 5.2/source/WSclient.cpp
@@ -6415,11 +6415,6 @@ void ReceiveLevelUp(const BYTE* ReceiveBuffer, int Size)
         CharacterAttribute->wMaxMinusPoint = Data->wMaxMinusPoint;
     }
 
-    wchar_t szText[256] = { NULL, };
-    uint64_t iExp = CharacterAttribute->NextExperience - CharacterAttribute->Experience;
-    swprintf(szText, GlobalText[486], iExp);
-    g_pSystemLogBox->AddText(szText, SEASON3B::TYPE_SYSTEM_MESSAGE);
-
     CharacterMachine->CalculateNextExperince();
 
     OBJECT* o = &Hero->Object;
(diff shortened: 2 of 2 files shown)


Source Main 5.2/source/WSclient.cpp
```diff
@@ -5329,7 +5329,7 @@ BOOL ReceiveDieExp(const BYTE* ReceiveBuffer, BOOL bEncrypted)
     }
     else
     {
-        g_pMainFrame->SetPreExp(CharacterAttribute->Experience);
```
nitoygo, line 5332: encounters exception due to data type size mismatch (?). please verify if this is the correct fix, or we need to change the data type of Experience to just DWORD instead

Source Main 5.2/source/WSclient.cpp
```diff
@@ -5417,7 +5417,7 @@ BOOL ReceiveDieExpLarge(const BYTE* ReceiveBuffer, BOOL bEncrypted)
...
```
nitoygo, line 5420: encounters exception due to data type size mismatch (?). please verify if this is the correct fix, or we need to change the data type of Experience to just DWORD instead

Source Main 5.2/source/WSclient.cpp
```diff
@@ -6416,7 +6416,7 @@ void ReceiveLevelUp(const BYTE* ReceiveBuffer, int Size)
     }
 
     wchar_t szText[256] = { NULL, };
-    WORD iExp = CharacterAttribute->NextExperience - CharacterAttribute->Experience;
+    uint64_t iExp = CharacterAttribute->NextExperience - CharacterAttribute->Experience;
```
nitoygo: Encounters exception when leveling up due to incorrect size

Source Main 5.2/source/WSclient.cpp
```diff
@@ -5329,7 +5329,7 @@ BOOL ReceiveDieExp(const BYTE* ReceiveBuffer, BOOL bEncrypted)
     }
     else
     {
-        g_pMainFrame->SetPreExp(CharacterAttribute->Experience);
+        g_pMainFrame->SetPreExp(CharacterAttribute->Experience & 0xFFFFFFFF);
```
nitoygo, line 5332: encounters exception due to data type size mismatch.

Source Main 5.2/source/WSclient.cpp
```diff
@@ -5417,7 +5417,7 @@ BOOL ReceiveDieExpLarge(const BYTE* ReceiveBuffer, BOOL bEncrypted)
...
```
nitoygo, line 5420: encounters exception due to data type size mismatch.

Source Main 5.2/source/NewUIMyInventory.cpp
```diff
@@ -328,8 +328,8 @@ bool CNewUIMyInventory::IsEquipable(int iIndex, ITEM* pItem)
 
     if (pItem->Type == ITEM_DARK_RAVEN_ITEM)
     {
-        const PET_INFO* pPetInfo = GetPetInfo(pItem);
-        const WORD wRequireCharisma = (185 + (pPetInfo->m_wLevel * 15));
+        auto pPetInfo = GetPetInfo(pItem);
+        auto wRequireCharisma = (185 + (pPetInfo->m_wLevel * 15));
```
nitoygo: WORD can't hold the computed charisma. Used auto instead.

sven-n: WORD can hold the required charisma. The pet level is usually capped at level 50, so the result is nowhere near 65k. The root of the issue is probably something else

nitoygo: in this code block it only checks whether the charisma is enough or not, returns boolean. i think this should enough for this particular code block?

Source Main 5.2/source/WSclient.cpp
sven-n: I think you can actually remove this 4 lines. There is no need to print out this exp gain, because the server sent a corresponding exp message before anyway.

Source Main 5.2/source/NewUIMyInventory.cpp
nitoygo: yeah, rethinking about that formula, i get the point that it should not exceed 65k

sven-n: The pet data is somehow not initialized:
grafik

sven-n: I think this is due the increased frame rate... I get this uninitialized data for two frames, until the pet info is received from the server. However, with this code, the intent should be clearer:
```
    if (pItem->Type == ITEM_DARK_RAVEN_ITEM)
    {
        const auto pPetInfo = GetPetInfo(pItem);
        if (pPetInfo->m_dwPetType == PET_TYPE_NONE)
        {
            return false;
        }

        const auto requiredCharisma = (185 + (pPetInfo->m_wLevel * 15));
        if (requiredCharisma > wCharisma)
        {
            return false;
        }
    }
```

nitoygo: I'll apply this suggestion. Many thanks!

nitoygo: Tho i still don't understand how the increased frame rate might have caused this

sven-n: The server is sending this data on-demand for non-equipped raven/horses. As soon as you hover over the item with your mouse pointer, it sends a packet which requests this info. I think that with the lower frame rate, the response from the server was fast enough to be available at the next frame.

sven-n: Related: https://github.com/sven-n/MuMain/issues/12

Source Main 5.2/source/WSclient.cpp
nitoygo: Agreed, I applied the suggestion and works fine.

Source Main 5.2/source/NewUIMyInventory.cpp
nitoygo: Applied in latest commit. Kindly check. Thanks!

nitoygo:
Addressed remaining comments. Thanks :)

sven-n (approved):

sven-n:
Thank you :)

PR #51, opened 2024-10-25
Improved Frustrum Check

nitoygo:


sven-n:
I'd like to understand the solution first, before I merge it 😉
Maybe we need to correct the calculation of the frustum to fix this better.

nitoygo:
For sure. :)

PR #54, opened 2024-10-30
MU Helper Implementation (Client Side)

nitoygo:
This aims to add the missing MU Helper implementation.  
The macro logic will be running in client side as to not overwhelm the server.  
If the need arises where the benefits of running it on the server side outweighs being in the client, we can move the logic there.  
For now, this should be a good usable starting point.


README.md
```diff
@@ -77,3 +76,4 @@ The client identifies itself with Version `2.04d` and serial `k1Pk2jcET48mxL3b`.
   * Louis
   * Qubit (tuservermu.com.ve)
   * Community members of RaGEZONE and tuservermu.com.ve for posting fixes
+  * [Nitoy](https://github.com/nitoygo) for the MU Helper
```
nitoygo, line 77: shameless plug :))

Source Main 5.2/source/GMHellas.cpp
```diff
@@ -740,10 +740,10 @@ void SetActionDestroy_Def(OBJECT* o)
 {
     if (o->Type != MODEL_PLAYER)
     {
-        if (g_isCharacterBuff(o, eBuff_PhysDefense))
+        if (g_isCharacterBuff(o, eBuff_WizDefense))
```
sven-n, line 743: Are you sure this is correct? In the following line it refers to `ACTION_DESTROY_PHY_DEF`.

Source Main 5.2/source/MUHelper/MuHelper.cpp
```diff
@@ -0,0 +1,1004 @@
...
+#include "NewUISystem.h"
+#include "Utilities/Log/muConsoleDebug.h"
+#include "SkillManager.h"
+#include "PartyManager.h"
+#include "MapManager.h"
+#include "WSclient.h"
+
+#include "MuHelper.h"
+
+#define MAX_ACTIONABLE_DISTANCE        10
```
sven-n: Can these be changed to constexpr?

Source Main 5.2/source/MUHelper/MuHelperData.h
```diff
@@ -0,0 +1,84 @@
+#pragma once
+
+#include <set>
+#include <array>
+#include <string>
+#include "WSclient.h"
+
+#define MUHELPER_ATTACK_ALWAYS                   0x00000000
```
sven-n: Flags enum? E.g. like `EPathNodeState`

Source Main 5.2/source/GMHellas.cpp
nitoygo: I just renamed the enum itself. I have verified that it is actually the soul barrier so i renamed it for less confusion.

Source Main 5.2/source/MUHelper/MuHelper.cpp
nitoygo: Updated in latest commit

Source Main 5.2/source/MUHelper/MuHelperData.h
nitoygo: Updated in latest commit

Source Main 5.2/source/GMHellas.cpp
nitoygo: Chances are, the line below is also misnamed. That or this is an inherent mistake from the old code.

nitoygo: I renamed the ACTION_DESTROY_PHY_DEF to ACTION_DESTROY_WIZ_DEF for consistency.

PR #68, opened 2024-11-24
Path Synchronization / MU Helper Related Fixes

nitoygo:
Updates:
- Object Moved Handler
  - Client now uses the path/steps data from the server
  - Character and monster moves are now 'synchronized' across multiple observers
  - Testing:
    - Login 2 characters (Debug > Start New Instance  2x)
    - Try moving 1 of the characters and notice the movements on the other window
    - Try moving 1 tile at a time, try moving more than 1 tile (verify that the movements on both clients are in sync)

- MU Helper: 
  - Using shorter range skills (vs target distance) now moves the character closer to the target before casting
    - (Limited to 2 steps at a time)
  - Path finding around obstacle for wizard char now works (previously, only non-wizard can do it)
  - Combo activation now needs all 3 skills to be selected (unselecting one unchecks the combo checkbox)

- Dependency Updates
  - Updated network/client library versions


sven-n (approved):

sven-n:
Thanks again :)

PR #69, opened 2024-11-27
Fixed no damage on Cursed Wizard when targeted with area skill

### LINKED ISSUE #539: Area Skills with explicit target do not damage specific monsters
**Describe the bug**
When casting an area skill like Twisting Slash or Rageful Blow by right clicking to a monster, the right clicked monster does not receive damage

**To Reproduce**
1. Login BK, go to a map (e.g. Lost Tower)
2. Select twisting slash or rageful blow as skill
3. Go close to a monster, right click monster to cast the skill
4. Notice that the right clicked monster does not have damage

**Expected behavior**
Area skills like Twisting Slash or Rageful Blow should deal damage to all monster within range (even if target is selected)

**Screenshots**
image

**Additional context**
- As shown in the screen shot, it damages other monsters, other than the one right clicked
- Happens in 6.3 client and 5.2 MuMain

nitoygo:
9f11723f7cb5719c72bc2fc1b97b40cb66211cc7 Fixes: https://github.com/MUnique/OpenMU/issues/539

Other fixes included in PR:
-Fixed issue where MU Helper targets are cleared when stopped then started, hero won't attack unless monsters move or attack.

### DIFF
--- a/Source Main 5.2/source/MUHelper/MuHelper.cpp
+++ b/Source Main 5.2/source/MUHelper/MuHelper.cpp
@@ -94,8 +94,6 @@ namespace MUHelper
 
         m_iSecondsElapsed = 0;
         m_iSecondsAway = 0;
-        m_setTargets.clear();
-        m_setTargetsAttacking.clear();
 
         m_bTimerActivatedBuffOngoing = false;
         m_bPetActivated = false;
--- a/Source Main 5.2/source/MUHelper/MuHelper.cpp
+++ b/Source Main 5.2/source/MUHelper/MuHelper.cpp
@@ -702,7 +702,8 @@ namespace MUHelper
                 {
                     m_iCurrentTarget = GetFarthestAttackingTarget();
                 }
-                else
+                
+                if (m_iCurrentTarget == -1)
                 {
                     m_iCurrentTarget = GetNearestTarget();
                 }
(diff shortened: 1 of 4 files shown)


Source Main 5.2/source/ZzzInterface.cpp
```diff
@@ -1855,7 +1855,7 @@ int	getTargetCharacterKey(CHARACTER* c, int selected)
                 return -1;
         }
     }
-    else if (sc->PK >= PVP_MURDERER2 || (HIBYTE(GetAsyncKeyState(VK_CONTROL)) == 128 && sc != Hero))
+    else if ((sc->PK >= PVP_MURDERER2 && sc->Object.Type == KIND_PLAYER) || (HIBYTE(GetAsyncKeyState(VK_CONTROL)) == 128 && sc != Hero))
```
nitoygo, line 1858: Monsters like 'Cursed Wizard' are set with PVP_MURDERER2 status. This check is intended for players.

Source Main 5.2/source/MUHelper/MuHelper.cpp
```diff
@@ -94,8 +94,6 @@ namespace MUHelper
 
         m_iSecondsElapsed = 0;
         m_iSecondsAway = 0;
-        m_setTargets.clear();
-        m_setTargetsAttacking.clear();
```
nitoygo, line 98: When targets are cleared after stopping->starting the MU Helper anew, the character waits for monsters to move/attack for them to be listed again as targets. So let's not clear them, these will eventually be cleared when the targets in the set are invalid or out of range.

nitoygo:
See https://github.com/sven-n/MuMain/pull/70 instead

sven-n:
Thanks :)

Issue #78, opened 2024-12-08
Skill Nova not work


sonniit92:
I left the code to fix the bug when using nova skill that I found in your code.

In **void Attack(CHARACTER* c)**

Remove this
if (MouseLButtonPush || MouseLButton)
{
		MouseLButtonPush = false;
		Success = true;
}
	
In **int ExecuteSkill(CHARACTER* c, int Skill, float Distance);**

iSkillIndex = g_pSkillList->GetSkillIndex(Skill); will return an invalid value for the skill and thus the skill cannot be used.

Fix by

if (Skill == AT_SKILL_BLAST_HELL_BEGIN)
{
	if (g_pSkillList->GetSkillIndex(AT_SKILL_BLAST_HELL) == -1)
	{
		return 0;
	}
	iSkillIndex = g_pSkillList->GetSkillIndex(AT_SKILL_BLAST_HELL);
}
else
{
	iSkillIndex = g_pSkillList->GetSkillIndex(Skill);
	if (iSkillIndex == -1)
	{
		return 0;
	}
}

nitoygo:
Nice. But, why not send a PR?

sonniit92:
> Nice. But, why not send a PR?

I don't use github much so there might be some functions I don't know about, it's a bug so I posted here.

nitoygo:
Fixed in latest OpenMU+MuMain. Kindly close issue if you were able to verify this to be already working. Thanks!

PR #79, opened 2024-12-09
Multiple wizard skills bugfixes

nitoygo:
- Fixed teleport bug after teleporting ally
  - Removed usage of 'Teleport' global variable
- Fixed nova skill not activating

### DIFF
--- a/Source Main 5.2/source/NewUIMainFrameWindow.cpp
+++ b/Source Main 5.2/source/NewUIMainFrameWindow.cpp
@@ -1832,6 +1832,12 @@ int SEASON3B::CNewUISkillList::GetHotKey(int iHotKey)
 
 int SEASON3B::CNewUISkillList::GetSkillIndex(int iSkillType)
 {
+    // special handling for skills with different skill id for the trigger
+    if (iSkillType == AT_SKILL_BLAST_HELL_BEGIN)
+    {
+        iSkillType = AT_SKILL_BLAST_HELL;
+    }
+
     int iReturn = -1;
     for (int i = 0; i < MAX_MAGIC; ++i)
     {
--- a/Source Main 5.2/source/WSclient.cpp
+++ b/Source Main 5.2/source/WSclient.cpp
@@ -218,8 +218,6 @@ int MoveCount = 0;
 
 int CurrentSkill = 0;
 
-bool Teleport = false;
-
 int BuyCost = 0;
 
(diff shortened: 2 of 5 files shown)


Source Main 5.2/source/WSclient.cpp
```diff
@@ -218,8 +218,6 @@ int MoveCount = 0;
 
 int CurrentSkill = 0;
 
-bool Teleport = false;
```
nitoygo, line 221: This is the root of all evil. Better to remove it. In the server side, this is tied to the player, while here, it is used for everything that is teleporting.

Source Main 5.2/source/ZzzInterface.cpp
```diff
@@ -6149,16 +6148,8 @@ void AttackWizard(CHARACTER* c, int Skill, float Distance)
...
-                    if (Teleport)
-                    {
-                        Teleport = false;
-                    }
-                    else
-                    {
-                        Teleport = true;
-                        SocketClient->ToGameServer()->SendTeleportTarget(tc->Key, TargetX, TargetY);
-                        SetPlayerTeleport(tc);
-                    }
```
nitoygo, line 6161: Removing the Teleport boolean checks here is the actual fix to the teleport bug, but I think it's better to remove the Teleport variable altogether. if ever it's needed, it's best to check o->Teleport value instead.

Source Main 5.2/source/NewUIMainFrameWindow.cpp
```diff
@@ -1832,6 +1832,12 @@ int SEASON3B::CNewUISkillList::GetHotKey(int iHotKey)
 
 int SEASON3B::CNewUISkillList::GetSkillIndex(int iSkillType)
 {
+    // special handling for skills with different skill id for the trigger
+    if (iSkillType == AT_SKILL_BLAST_HELL_BEGIN)
+    {
+        iSkillType = AT_SKILL_BLAST_HELL;
+    }
```
nitoygo, line 1839: Taking inspiration from the comment here:
https://github.com/sven-n/MuMain/issues/78#issue-2725294683
but with a slightly different approach

Source Main 5.2/source/ZzzInterface.cpp
sven-n: Yes, global variables are a bad idea anyway 👍

sven-n (approved):

PR #81, opened 2024-12-11
Use std::chrono for timer; Add yield in main scene loop spin phase

nitoygo:
I did not measure any performance differences. This is just to slowly migrate to std libraries instead of windows-specific.

### DIFF
--- a/Source Main 5.2/source/SpinLock.h
+++ b/Source Main 5.2/source/SpinLock.h
@@ -1,5 +1,6 @@
 #pragma once
 #include <atomic>
+#include <thread>
 
 class SpinLock {
     std::atomic_flag locked = ATOMIC_FLAG_INIT;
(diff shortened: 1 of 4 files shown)


Source Main 5.2/source/ZzzScene.cpp
```diff
@@ -2498,6 +2499,11 @@ void MainScene(HDC hDC)
                 {
                     break;
                 }
+                // reduce CPU usage by yielding during spin phase
+                if (spin_ms - spinned_ms > spin_yield_threshold_ms)
+                {
+                    std::this_thread::yield();
+                }
```
nitoygo, line 2506: I think this is necessary for cases where the sleep threshold was not reached, to prevent busy-spinning for almost 8ms

Source Main 5.2/source/SpinLock.h
```diff
@@ -8,7 +9,7 @@ class SpinLock {
     {
         while (locked.test_and_set(std::memory_order_acquire))
         {
-            Sleep(0);
+            std::this_thread::yield();
```
nitoygo, line 12: Use cross platform mechanism instead of windows Sleep()

PR #82, opened 2024-12-12
Fixed deathstab visual & fps bugs

nitoygo:
* Fixed FPS drop when casting deathstab
* Fixed incorrect deathstab animation


Source Main 5.2/source/w_ObjectInfo.h
```diff
@@ -141,7 +141,7 @@ class OBJECT
     BYTE          Kind;
     WORD		Skill;
     BYTE		  m_byNumCloth;
-    BYTE		  m_byHurtByOneToOne;
+    BYTE		  m_byHurtByDeathstab;
```
sven-n: float?

Source Main 5.2/source/ZzzCharacter.cpp
```diff
@@ -3817,7 +3817,7 @@ void MoveCharacter(CHARACTER* c, OBJECT* o)
                 }
             }
         }
-        o->m_byHurtByOneToOne--;
+        o->m_byHurtByDeathstab--;
```
sven-n: -= FPS_ANIMATION_FACTOR ?

sven-n (commented):
Thanks again. What do you think... shouldn't the `m_byHurtByDeathstab` be a float now? Because previously, it counted down per frame, now it should consider the fps

Source Main 5.2/source/w_ObjectInfo.h
nitoygo: i haven't thought of that. let me check...

nitoygo: This is the part where it's set,  
`to->m_byHurtByDeathstab = 35;` 
Does this mean it's meant to stay for 35 frames? but now it should stay for the corresponding animation factor?

Source Main 5.2/source/GM_Raklion.cpp
```diff
@@ -1905,7 +1905,7 @@ bool CGM_Raklion::RenderMonsterVisual(CHARACTER* c, OBJECT* o, BMD* b)
                         OBJECT* to = &tc->Object;
                         if (10 <= c->AttackTime && to->Live)
                         {
-                            to->m_byHurtByOneToOne = 35;
+                            to->m_byHurtByDeathstab = 35;
```
nitoygo, line 1908: hey @sven-n, do we also need to update this initial value based on the fps animation factor? or is this fine, and we only need to update the part where it's decremented?

Source Main 5.2/source/w_ObjectInfo.h
nitoygo: I applied the suggestion, animation and duration seems fine.

Source Main 5.2/source/ZzzEffectJoint.cpp
```diff
@@ -4235,7 +4235,7 @@ void MoveJoint(JOINT* o, int iIndex)
 
             vec3_t MagicPos;
             GetMagicScrew(iIndex * 17721, MagicPos, 1.4f);
-            VectorScale(MagicPos, 300.0f, MagicPos);
+            //VectorScale(MagicPos, 300.0f, MagicPos);
```
nitoygo, line 4238: Removing this line fixed the weird deathstab magic animation

PR #88, opened 2024-12-15
Fix for Unpack jewels (Lahap) not working

### LINKED ISSUE #84: Unpacking jewels does not work -jkok25
Packing Jewels work correctly (according to jkok25)

nitoygo:
Fixes https://github.com/sven-n/MuMain/issues/84

### DIFF
--- a/Source Main 5.2/source/CComGem.cpp
+++ b/Source Main 5.2/source/CComGem.cpp
@@ -34,7 +34,6 @@ namespace COMGEM
 
 void COMGEM::SendReqUnMix()
 {
-    iUnMixIndex += MAX_EQUIPMENT;
     SocketClient->ToGameServer()->SendLahapJewelMixRequest(0x01, m_cGemType / 2, iUnMixLevel, iUnMixIndex);
 }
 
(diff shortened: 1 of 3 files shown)


Source Main 5.2/source/CComGem.cpp
```diff
@@ -34,7 +34,7 @@ namespace COMGEM
 
 void COMGEM::SendReqUnMix()
 {
-    iUnMixIndex += 12;
+    iUnMixIndex += MAX_EQUIPMENT;
```
nitoygo, line 37: Not sure if this is better, or if adding the entry to the list of candidates (packed jewels) should already have this offset added.

Source Main 5.2/source/CComGem.cpp
```diff
@@ -34,7 +34,6 @@ namespace COMGEM
 
 void COMGEM::SendReqUnMix()
 {
-    iUnMixIndex += 12;
```
nitoygo, line 37: No need to add this offset since we already added it when populating the list (for the dialog)

Source Main 5.2/source/CComGem.cpp
```diff
@@ -76,7 +75,7 @@ bool COMGEM::FindWantedList()
         if (isCompiledGem(pItem))
         {
             INTBYTEPAIR p;
-            p.first = pItem->y * pNewInventoryCtrl->GetNumberOfColumn() + pItem->x;
+            p.first = pItem->y * pNewInventoryCtrl->GetNumberOfColumn() + pItem->x + MAX_EQUIPMENT_INDEX;
```
nitoygo, line 78: Include the offset to prevent invalid access when performing the unmix

sven-n:
Thanks again :)

PR #91, opened 2024-12-17
Workaround for incorrect harmony option display

### LINKED ISSUE #80: Jewel of Harmony Options not rendered correctly
Probably related to char* -> wchar_t* migration...

image

nitoygo:
Fixes https://github.com/sven-n/MuMain/issues/80

image 
image

Note:  
I think the permanent fix would be to update JewelOfHarmonyOption_eng.bmd to use wchar_t[] for the names. If in case this bmd file is updated, we need to undo this patch.


sven-n:
Thanks for the fix :)

PR #92, opened 2024-12-20
Fixed string rendering in goblin dialogs

nitoygo:
Fixed gibberish string displayed in the assembly recipe.  
Below is an example in Chaos Goblin (this bug is also present in Aida1 goblins)

Before:
image

After:
image

### DIFF
--- a/Source Main 5.2/source/NewUIMixInventory.cpp
+++ b/Source Main 5.2/source/NewUIMixInventory.cpp
@@ -525,7 +525,7 @@ void CNewUIMixInventory::RenderFrame()
         g_pRenderText->SetTextColor(220, 220, 220, 255);
         g_pRenderText->SetBgColor(40, 40, 40, 128);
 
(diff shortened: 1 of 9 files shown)


Source Main 5.2/source/NewUIMixInventory.cpp
```diff
@@ -525,7 +525,7 @@ void CNewUIMixInventory::RenderFrame()
         g_pRenderText->SetTextColor(220, 220, 220, 255);
         g_pRenderText->SetBgColor(40, 40, 40, 128);
 
-        wchar_t szTempText[2][100];
+        wchar_t szTempText[2][100] = { 0 };
```
nitoygo, line 528: This is the main fix for this PR, this resolves the text rendering issues with goblin (item mix) NPCs.  

Other diffs here are just to lessen the compile warnings related to some initialization and string usage

PR #105, opened 2024-12-27
Make sure only monster/player are selected as targets for MU Helper

### LINKED ISSUE #104: MU Helper Stuck when a non-attackable enemy is nearby
Enemy objects like this canon confuses MU Helper:
image

Expected behavior:
Such enemies should not be included in the MU Helper targets since it's not attackable

nitoygo:
Fixes: https://github.com/sven-n/MuMain/issues/104

### DIFF
--- a/Source Main 5.2/source/WSclient.cpp
+++ b/Source Main 5.2/source/WSclient.cpp
@@ -3216,7 +3216,10 @@ void ReceiveAttackDamageExtended(const BYTE* ReceiveBuffer)
     auto ShieldDamage = Data->ShieldDamage;
 
     g_ConsoleDebug->Write(MCD_RECEIVE, L"0x15 [ReceiveAttackDamageExtended(%d %d)]", AttackPlayer, Damage);
-    MUHelper::g_MuHelper.AddTarget(Key, true);
+    if (IsMonster(c) || IsPlayer(c))
+    {
+        MUHelper::g_MuHelper.AddTarget(Key, true);
+    }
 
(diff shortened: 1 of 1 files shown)


Source Main 5.2/source/WSclient.cpp
```diff
@@ -3216,7 +3216,10 @@ void ReceiveAttackDamageExtended(const BYTE* ReceiveBuffer)
     auto ShieldDamage = Data->ShieldDamage;
 
     g_ConsoleDebug->Write(MCD_RECEIVE, L"0x15 [ReceiveAttackDamageExtended(%d %d)]", AttackPlayer, Damage);
-    MUHelper::g_MuHelper.AddTarget(Key, true);
+    if (IsMonster(c) || IsPlayer(c))
```
nitoygo, line 3219: Prevents object.Kind == KIND_TRAP from being added as target

Source Main 5.2/source/WSclient.cpp
```diff
@@ -3295,7 +3298,10 @@ void ReceiveAction(const BYTE* ReceiveBuffer, int Size)
         c->Object.AnimationFrame = 0;
 
         c->TargetCharacter = HeroIndex;
-        MUHelper::g_MuHelper.AddTarget(Key, true);
+        if (IsMonster(c) || IsPlayer(c))
```
nitoygo, line 3301: Prevents object.Kind == KIND_TRAP from being added as target

Issue #102, opened 2024-12-27
Incorrect Skins


nitoygo:
Some equipments show parts of the body:
![image](https://github.com/user-attachments/assets/9b0b73d9-5d32-4e75-a06b-e0f248d622e7)

Weird trees in tarkan, weird texture of some accessories:
![image](https://github.com/user-attachments/assets/97083dd9-0aa3-41ca-8965-96d9ca2c104f)

nitoygo:
Not sure if related to this (https://github.com/sven-n/MuMain/pull/64/commits/887b527ee0c3417ad33ba5afe6abdb453d98135c), but this might have been introduced after the skins fix update

sven-n:
It's a bit hard to debug these kind of errors. As I don't have client file editing tools, can you tell me which files (of Data/Object9/) are involved for the trees in tarkan?
I'll then try to track down the issue with the right breakpoints :)

fantasticgs:
Object32.bmd Object33.bmd Object34.bmd Three tree models

sven-n:
I think I found the issue. I'll try to explain:

When loading a map, the code loads models and textures based on common names, e.g. from `Data\Object9\ObjectX.bmd` in a loop at https://github.com/sven-n/MuMain/blob/f4a5aefd3c04e7c8338e8d8f2c36038baf91f97f/Source%20Main%205.2/source/MapManager.cpp#L1153-L1161 and their textures at https://github.com/sven-n/MuMain/blob/f4a5aefd3c04e7c8338e8d8f2c36038baf91f97f/Source%20Main%205.2/source/LoadData.cpp#L76-L80

So far so good. However, when the client loads one texture with a specific name in the LoadData::OpenTexture, it will keep it and not reload it again. Example: Client loads map of Folder `Data\Object74` and a texture `Data\Object74\test2.jpg`. Then it enters another map which has texture `Data\Object9\test2.jpg`, it will not reload it, but uses the wrong texture of the old map.

PR #115, opened 2025-01-02
Fixed texture loading for world objects

sven-n:
For example, it fixes the tree texture issue of #102.
The issues with some armors in the inventory is still there.


nitoygo:
Additionally, the rings/pendant skin issue was also fixed by this commit. :)

PR #124, opened 2025-01-06
Prevent targeting dead monsters

nitoygo:


Source Main 5.2/source/ZzzInterface.cpp
```diff
@@ -1575,6 +1575,11 @@ bool CheckAttack()
 
     CHARACTER* c = &CharactersClient[SelectedCharacter];
 
+    if (c->Dead)
+    {
+        return false;
+    }
```
nitoygo, line 1581: This will prevent dead monsters from being selected as targets again.

Source Main 5.2/source/ZzzScene.cpp
```diff
@@ -2465,6 +2464,8 @@ void MainScene(HDC hDC)
             SwapBuffers(hDC);
         }
 
+        g_render_lock->unlock();
```
nitoygo, line 2467: Before entering the sleep/spin phase of FPS synchronization, we allow other threads to alter the game state.

Source Main 5.2/source/ZzzScene.cpp
```diff
@@ -2509,6 +2510,8 @@ void MainScene(HDC hDC)
             current_tick_count += static_cast<int>(rest_ms);
         }
 
+        g_render_lock->lock();
```
nitoygo, line 2513: Then lock it again for the next rendering iteration to be data integrity safe

PR #125, opened 2025-01-06
Fixed death animation for monsters (e.g. bloody wolf in tarkan)

sven-n:


nitoygo (approved):
Works on my machine :)

PR #131, opened 2025-01-08
Fix other inventory issues

### LINKED ISSUE #128: Client crashes when closing inventory with item clicked
**Describe the bug**
When an item is held (clicked by cursor and not yet placed anywhere), and you press 'v' to close the inventory, the client crashes

**Steps to Reproduce**
1. Open inventory
2. Click any item (don't drop or move it somewhere yet)
3. Press 'v' to close the inventory
4. See exception happens:
image

### LINKED ISSUE #130: Cannot deposit items to vault when inventory extension is open
**Describe the bug**
Cannot deposit items to vault when inventory extension is open

**Steps to Reproduce**
1. Login a character with storage extension
2. Go to vault
3. Open extension
4. Transfer an item from inventory to vault (using left clicks)
// Item cannot be transferred

**Note**
This only happens with manual transfer using left clicks, right clicking the item works fine

nitoygo:
Fixes: https://github.com/sven-n/MuMain/issues/130
Fixes: https://github.com/sven-n/MuMain/issues/128

### DIFF
--- a/Source Main 5.2/source/NewUIManager.cpp
+++ b/Source Main 5.2/source/NewUIManager.cpp
@@ -20,16 +20,22 @@ SEASON3B::CNewUIManager::~CNewUIManager()
 
 void SEASON3B::CNewUIManager::AddUIObj(DWORD dwKey, CNewUIObj* pUIObj)
 {
+    std::unique_lock<std::shared_mutex> lock(m_mutex);
+
     auto mi = m_mapUI.find(dwKey);
     if (mi == m_mapUI.end())
     {
         m_vecUI.push_back(pUIObj);
         m_mapUI.insert(type_map_uibase::value_type(dwKey, pUIObj));
     }
+
+    std::sort(m_vecUI.begin(), m_vecUI.end(), CompareKeyEventOrder);
 }
 
 void SEASON3B::CNewUIManager::RemoveUIObj(DWORD dwKey)
 {
+    std::unique_lock<std::shared_mutex> lock(m_mutex);
+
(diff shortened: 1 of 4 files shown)


Source Main 5.2/source/NewUIManager.cpp
```diff
@@ -44,6 +50,8 @@ void SEASON3B::CNewUIManager::RemoveUIObj(DWORD dwKey)
 
 void SEASON3B::CNewUIManager::RemoveUIObj(CNewUIObj* pUIObj)
 {
+    std::unique_lock<std::shared_mutex> lock(m_mutex);
```
nitoygo, line 53: This will act as a write lock to exclusively alter m_vecUI.

Source Main 5.2/source/NewUIManager.cpp
```diff
@@ -141,11 +157,16 @@ bool SEASON3B::CNewUIManager::UpdateMouseEvent()
 
 bool SEASON3B::CNewUIManager::UpdateKeyEvent()
 {
+    type_vector_uibase vecUI;
+    {
+        std::shared_lock<std::shared_mutex> lock(m_mutex);
+        vecUI = m_vecUI;
+    }
```
nitoygo, line 164: While this will act as a read lock to be able to read m_vecUI without getting blocked.  
Since the AddUIObj and RemoveUIObj might alter the vector in parallel, we will use a copy here that is unaffected by these alters.  

The exception happened when the entry removed during RemoveUIObj was dereferenced in this function.  Using the copied vector prevents this error.

PR #135, opened 2025-01-10
Moved frame limiting logic to main loop

nitoygo:


Source Main 5.2/source/ZzzScene.cpp
```diff
@@ -2803,10 +2743,30 @@ float g_Luminosity;
...
+    }
+
+    return false;
+}
+
 void Scene(HDC hDC)
 {
+    if (!CheckRenderNextScene())
+    {
+        return;
```
sven-n, line 2764: Why return here?

PR #138, opened 2025-01-11
FPS Limiter x Packet handler delay

nitoygo:


Source Main 5.2/source/ZzzScene.cpp
```diff
@@ -2803,44 +2743,67 @@ float g_Luminosity;
...
-    {
+
+        last_render_tick_count = g_pTimer->GetTimeElapsed();
     }
 
     wglMakeCurrent(nullptr, nullptr);
     g_render_lock->unlock();
+
+    // allow context switching for network thread
+    std::this_thread::yield();
```
nitoygo, line 2806: This should allow context switching, without the need for complex sleep/spin logic

Source Main 5.2/source/ZzzScene.cpp
```diff
@@ -2803,44 +2743,67 @@ float g_Luminosity;
...
 {
     g_render_lock->lock();
     wglMakeCurrent(hDC, g_hRC);
-    try
+
+    if (CheckRenderNextScene())
     {
-        g_Luminosity = sinf(WorldTime * 0.004f) * 0.15f + 0.6f;
-        switch (SceneFlag)
+        last_render_tick_count = g_pTimer->GetTimeElapsed();
```
nitoygo, line 2767: Moving it here fixed the incorrect FPS calculation

PR #139, opened 2025-01-11
Moved frame limiter logic out of the render lock.

sven-n:
Now, the game can process network packets (and other stuff in the future) during these delays.


sven-n:
Closed, #140 is the better approach

PR #140, opened 2025-01-11
FPS Limiter fixed

nitoygo:
Updates https://github.com/sven-n/MuMain/pull/135

Major changes:  
- No more locking  
- Packets are immediately queued to main loop without waiting for lock
- No more opengl context switching
- Moved sleep phase outside Scene() to not block main loop
- Sleep only allowed when there is no queued message


Source Main 5.2/source/ZzzScene.cpp
```diff
@@ -2803,44 +2743,67 @@ float g_Luminosity;
...
 {
     g_render_lock->lock();
     wglMakeCurrent(hDC, g_hRC);
-    try
+
+    if (CheckRenderNextScene())
     {
-        g_Luminosity = sinf(WorldTime * 0.004f) * 0.15f + 0.6f;
-        switch (SceneFlag)
+        last_render_tick_count = g_pTimer->GetTimeElapsed();
```
nitoygo, line 2767: Placing this here fixes the incorrect FPS calculation.

Source Main 5.2/source/ZzzScene.cpp
```diff
@@ -2803,44 +2743,67 @@ float g_Luminosity;
...
     }
-    catch (const std::exception&)
-    {
-    }
 
     wglMakeCurrent(nullptr, nullptr);
     g_render_lock->unlock();
+
+    // allow context switching for network thread
+    std::this_thread::yield();
```
nitoygo, line 2806: This yield should prevent packet processing to be delayed.

Source Main 5.2/source/Winmain.cpp
```diff
@@ -1272,6 +1271,15 @@ MSG MainLoop()
         }
         else
         {
+            if (EnableSocket && (SocketClient == nullptr || !SocketClient->IsConnected()))
+            {
+                g_ErrorReport.Write(L"> Connection closed. ");
+                g_ErrorReport.WriteCurrentTime();
+                g_ConsoleDebug->Write(MCD_NORMAL, L"Connection closed");
+                MessageBox(NULL, L"Connection closed", L"Error", MB_OK);
```
nitoygo, line 1279: About this disconnection notice, you may change it back to in-game dialog.

Source Main 5.2/source/SpinLock.h
```diff
@@ -9,7 +9,6 @@ class SpinLock {
     {
         while (locked.test_and_set(std::memory_order_acquire))
         {
-            std::this_thread::yield();
```
nitoygo, line 12: Prevents unnecessary context switching - that might cause a contesting thread to miss a turn.

Source Main 5.2/source/WSclient.cpp
```diff
@@ -14553,19 +14555,25 @@ static void HandleIncomingPacket(int32_t Handle, const BYTE* ReceiveBuffer, int3
...
+}
+
+static void HandleIncomingPacket(int32_t Handle, const BYTE* ReceiveBuffer, int32_t Size)
+{
+    auto Packet = std::make_unique<PacketInfo>();
+    Packet->ReceiveBuffer = std::make_unique<BYTE[]>(Size);
+    std::copy(ReceiveBuffer, ReceiveBuffer + Size, Packet->ReceiveBuffer.get());
+    Packet->Size = Size;
+
+    PostMessage(g_hWnd, WM_RECEIVE_BUFFER, reinterpret_cast<WPARAM>(Packet.release()), 0);
```
nitoygo, line 14576: Received packet are immediately queued to the next game loop cycle (in the main thread) without waiting for render lock. Also, there's no longer a need to switch opengl context between threads.

Source Main 5.2/source/ZzzScene.cpp
```diff
@@ -2803,44 +2756,61 @@ float g_Luminosity;
...
+        return true;
+    }
+
+    return false;
+}
+
 void Scene(HDC hDC)
 {
-    g_render_lock->lock();
-    wglMakeCurrent(hDC, g_hRC);
```
nitoygo, line 2809: In the Scene() method, there's also no need for renderer lock and context switching of opengl.

Source Main 5.2/source/SpinLock.h
nitoygo: With the latest commit, this SpinLock is no longer used since there is no locking needed.

Source Main 5.2/source/Winmain.cpp
```diff
@@ -562,6 +561,11 @@ LONG FAR PASCAL WndProc(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lParam)
             break;
         }
         break;
+    case WM_RECEIVE_BUFFER: {
+        auto Packet = std::unique_ptr<PacketInfo>(reinterpret_cast<PacketInfo*>(wParam));
+        ProcessPacketCallback(Packet.release());
```
nitoygo, line 566: This is in the main thread and will be processed immediately when renderer is resting (during frame limiting)

Source Main 5.2/source/WSclient.cpp
```diff
@@ -14553,19 +14555,25 @@ static void HandleIncomingPacket(int32_t Handle, const BYTE* ReceiveBuffer, int3
     //return ( TRUE);
 }
 
-static void HandleIncomingPacketLocked(int32_t Handle, const BYTE* ReceiveBuffer, int32_t Size)
+void ProcessPacketCallback(const PacketInfo* Packet)
 {
-    g_render_lock->lock();
-    wglMakeCurrent(g_hDC, g_hRC);
```
nitoygo, line 14559: No locking needed since this is now in the same thread as the main loop.

Source Main 5.2/source/Winmain.cpp
```diff
@@ -1258,25 +1262,33 @@ bool ExceptionCallback(_EXCEPTION_POINTERS* pExceptionInfo)
 MSG MainLoop()
 {
     MSG msg;
+
+    constexpr auto target_resolution = 1;
+    auto precise = timeBeginPeriod(target_resolution);
```
nitoygo, line 1267: This increases the precision of sleep functions

sven-n:
Works like a charm, thanks for all the work! 👍

PR #144, opened 2025-01-15
Customizable number of messages processed per cycle

nitoygo:
- Let's user set the max events processed per cycle using `$winmsg` command
  - -1 is same as the old behavior where all events are processed first before rendering
  - adjust as needed (recommended: "`$winmsg 10`")
- Added average CPU usage info in the debug text next to FPS info
- Setting `$fps` to a value >= refresh rate when vsync is On sets the `target_fps` global variable back to -1

### DIFF
--- a/Source Main 5.2/source/Winmain.cpp
+++ b/Source Main 5.2/source/Winmain.cpp
@@ -35,6 +35,7 @@
 #include "./ExternalObject/leaf/ExceptionHandler.h"
 #include "./Utilities/Dump/CrashReporter.h"
 #include "./Utilities/Log/muConsoleDebug.h"
+#include "./Utilities/CpuUsage.h"
 #include "ProtocolSend.h"
 #include "ProtectSysKey.h"
 #include "MUHelper/MuHelper.h"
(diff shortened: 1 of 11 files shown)


Source Main 5.2/source/Winmain.cpp
```diff
@@ -1259,6 +1260,60 @@ bool ExceptionCallback(_EXCEPTION_POINTERS* pExceptionInfo)
...
+        {
+            numFilled++;
+        }
+
+        // Calculate the current average
+        currentAvg = sum / numFilled;
+
+        // Update the CPU_AVG every 250 ms
+        auto currentTime = std::chrono::steady_clock::now();
+        if (std::chrono::duration_cast<std::chrono::milliseconds>(currentTime - lastUpdateTime).count() >= 250)
```
nitoygo, line 1300: To view the changes closer to real-time, decrease this time interval.

Source Main 5.2/source/ZzzScene.cpp
```diff
@@ -2261,13 +2261,18 @@ extern int  GrabScreen;
...
-float ms_per_frame = 1000.f / target_fps;
+double target_fps = 60;
+double ms_per_frame = 1000.0 / target_fps;
 
-void SetTargetFps(float targetFps)
+void SetTargetFps(double targetFps)
 {
+    if (IsVSyncEnabled() && targetFps >= GetFPSLimit())
+    {
+        targetFps = -1;
```
nitoygo, line 2271: This diff is for the rendering to still behave as if it's still unlimited frames, instead of manually setting the FPS which would be handled manually by the fps limiter.

PR #150, opened 2025-02-11
Fixed Life Swell Effect and more

sven-n:
The effect was a bit too strong, too many rays came out of the ground. The joint effect itself spawns a bunch of additional joint effects during each move/frame. We must consider the higher fps there, otherwise too many effects are created.

The same issue was fixed for some other effects as well.


nitoygo:
Works! It was more obvious when tested with vsync off or with just high FPS.

nitoygo (approved):

PR #151, opened 2025-02-18
Fixed skin of fairy elf

sven-n:
The texture specified in the BMD is wrong. To fix this, I replaced the wrong texture file 'skin_warrior_01.ozj' with 'skin_archer_01.ozj'. Additionally, I did some code simplifications.


nitoygo (approved):
Nice, elf no longer a rock star hairstyle xD

PR #152, opened 2025-03-02
Adjust skill distance check in MU Helper

nitoygo:


Source Main 5.2/source/MUHelper/MuHelper.cpp
```diff
@@ -877,10 +877,11 @@ namespace MUHelper
 
             TargetX = (int)(pTarget->Object.Position[0] / TERRAIN_SCALE);
             TargetY = (int)(pTarget->Object.Position[1] / TERRAIN_SCALE);
+            int iDistance = ComputeDistanceBetween({ Hero->PositionX, Hero->PositionY }, { TargetX, TargetY });
 
             PATH_t tempPath;
             bool bHasPath = PathFinding2(Hero->PositionX, Hero->PositionY, TargetX, TargetY, &tempPath, m_iHuntingDistance + fSkillDistance);
-            bool bTargetNear = CheckTile(Hero, &Hero->Object, fSkillDistance);
+            bool bTargetNear = CheckTile(Hero, &Hero->Object, iDistance - fSkillDistance);
```
sven-n, line 884: Why is the distance calculated here, too? In `CheckTile`, the distance between the player and the target is calculated as well:
```
bool CheckTile(CHARACTER* c, OBJECT* o, float Range)
{
    if (!c || !o) return false;

    float dx = o->Position[0] - (TargetX * TERRAIN_SCALE + TERRAIN_SCALE * 0.5f);
    float dy = o->Position[1] - (TargetY * TERRAIN_SCALE + TERRAIN_SCALE * 0.5f);

    // Compare squared distance with squared range to avoid sqrt calculation
    float squaredDistance = dx * dx + dy * dy;
    float squaredRange = (TERRAIN_SCALE * Range) * (TERRAIN_SCALE * Range);

    return squaredDistance <= squaredRange;
}
```

If `iDistance` is lower than the `fSkillDistance`, the passed range is negative, isn't it?

nitoygo: You're right. I forgot about the behavior of CheckTile.
I'll be closing this ticket.

PR #162, opened 2025-05-13
Fix invalid LangSelection entry in the registry

yangyitzzg:
Currently, reading the LangSelection entry in the registry does not take effect and always defaults to "Eng" because the REG_SZ length is not calculated correctly.


sven-n:
Thanks :)

PR #165, opened 2025-05-23
Code cleanup by REGZ.PL

sven-n:


sven-n:
1. REGZ.PL is a user who sent me his changes.
2. The dependencies were removed, because they are not in use and probably will not be used. Additionally, they're outdated. So when they're needed again, they'll be added in their newest version.
3. Themida wasn't used as well.
4. I will add some entries to gitignore for IDE files in the future. For now it's okay to include these changes.

PR #169, opened 2025-05-28
2nd code cleanup by REGZ.PL

sven-n:


Source Main 5.2/Main.vcxproj
```diff
@@ -164,6 +164,7 @@
       <DataExecutionPrevention />
       <TargetMachine>MachineX86</TargetMachine>
       <ImageHasSafeExceptionHandlers>false</ImageHasSafeExceptionHandlers>
+      <Profile>true</Profile>
```
sven-n, line 167: It allows to performance profile the created code. I'm not sure if that makes sense for the debug build, but I'll leave that in for now.

Source Main 5.2/source/CBTMessageBox.h
sven-n, line 1: Now, pragma once is missing... but I think we can even remove the whole CCBTMessageBox class, it's not in use.

PR #183, opened 2025-06-14
Updated MU Helper skill list layout

nitoygo:
I updated the layout to be relative to the skill slots in the mu helper window.  
Also, I restored the behavior where the skills are added top/bottom from the center depending on the number of skills.
image
This was the original intention of the old code (but was buggy when the column is more than 1)


sven-n:
Thanks :)

PR #185, opened 2025-06-14
Fix for running Client on Linux with Proton

JPZV:
This fix two crashes that occurs when trying to run the game on Linux with Proton (specially with Steam Deck)

The first one is related to the game in general, where the `vswprintf` crashes whenever the game is started with no error message. This was fixed by changing it with its cousin `wvsprintf`

The second one is only related when building using the Debug profile (Or enabling the Debug Console). This was because of Handle being null and no child was enumerated while starting the Debug Console. This was fixed by removing that dependency and just checking if the Handle is null when needed. The side effect, at least on Linux, is that in some times there could be more than one console, yet I'm not sure if this is a error of the workaround or if this was made by proton itself.

This fix was tested in both Windows 11 and in Steam Deck with Proton 10 (Beta), both OS using Debug and Release profiles.

### DIFF
--- a/Source Main 5.2/source/Utilities/Log/ErrorReport.cpp
+++ b/Source Main 5.2/source/Utilities/Log/ErrorReport.cpp
@@ -129,7 +129,7 @@ void CErrorReport::Write(const wchar_t* lpszFormat, ...)
     wchar_t lpszBuffer[1024] = { 0, };
     va_list va;
     va_start(va, lpszFormat);
-    wvsprintf(lpszBuffer, lpszFormat, va);
+    vswprintf(lpszBuffer, 1024, lpszFormat, va);
     va_end(va);
 
     WriteDebugInfoStr(lpszBuffer);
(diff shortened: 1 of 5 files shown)


nitoygo:
Hey, just curious, how is your experience playing this MU client on steam deck?

JPZV:
It works very good actually. I play using the handheld mode (and from Game Mode itself, rather than the Desktop mode), and no issue other than the same as if I were on Windows (For example, I can only use Potions. Mana and Antidote doesn't appears on the action bar. The same issue appears on Windows too).

The installation was pretty rough tho. I had to add the Main.exe to Steam, then change the .exe to the installer of .Net Core for the Launcher, and then change the .exe to the Launcher itself. But after that it's just pressing Play, changing the gamepad layout (Pretty simple, specially because of the mouse pad of the Steam Deck), and that is.

There's one minor issue where after and hour or so the game starts to lags (from 60FPS to 30-45FPS), it's like if it has a memory leak somewhere, although according to the OS, the CPU, GPU and RAM doesn't changes that much. After restarting the game it fixes itself. I'm not sure if that happens on Windows because I'm capped to 30FPS (because I'm using RDP)

nitoygo:
Wow, that lag after an hour seems weird. I'm not sure if I experience that in PC. I wonder if that's a proton issue

sven-n:
Thanks :)

PR #189, opened 2025-06-17
Update ErrorReport.cpp

dracy21:
Advantages:
✅ Safe: The old file is always deleted before a new one is created.
✅ Efficient: No more checking the size or content of the previous file.
✅ Clean: Old logs don't accumulate over time.

This approach is ideal if you want to start with a fresh log every time the application runs. 🚀

✅ Full Header Compliance:
Implemented all methods declared in the header file
Matched all function signatures exactly 

✅ Fixed Linking Errors:
Added proper DirectSound and IMM32 library linking
Implemented all previously missing methods

✅ Robust Error Handling:
Added comprehensive checks for all API calls
Safe string handling throughout

✅ Memory Safety:
Used secure string functions (_s versions)
Proper buffer size management

✅ Complete Feature Set:
Full system information reporting
OpenGL capabilities
Sound card detection
IME/Keyboard information

Is not complet yet but is god
At least it's organized and you can understand something.

### DIFF
--- a/Source Main 5.2/source/Utilities/Log/ErrorReport.cpp
+++ b/Source Main 5.2/source/Utilities/Log/ErrorReport.cpp
@@ -1,5 +1,5 @@
 // ErrorReport.cpp: implementation of the CErrorReport class.
-//
+// By Dr@Cy ...
 //////////////////////////////////////////////////////////////////////
 
 #include "stdafx.h"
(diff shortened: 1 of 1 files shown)


Mosch0512:
i would not create a log for each start but rather for each new day.

PR #190, opened 2025-06-17
Update ErrorReport.h

dracy21:
At least it's organized and you can understand something.
------------------------VERsIOn.2.0🚀 ----------------------------


sven-n:
Why is there implementation code in the header file?
And why are there two pull requests (#190 and #189), one for cpp and one for the header? Do they somehow belong together?

dracy21:
PR #189 was incomplete — the .cpp file depends on the .h file from PR #190 , so they belong together. The implementation won’t work without the corresponding header.

PR #191, opened 2025-06-17
Update ZzzOpenData.cpp

dracy21:
Now, when OpenMonsterModel(153) appears, in MuError.log it will be replaced with the monster's actual name from the enum (SPIDER_EGGS_1, ICE_GIANT, etc.), making the logs much easier to read. 🚀

default: return L"UNKNOWN";


sven-n (changes requested):
Hi, thanks, that's a good idea. However, it has two points for improvement IMHO:
1. It's incomplete, not all model names are mapped.
2.  Instead "UNKNOWN", we could still print out the `Type` value. 

And just a hint... It's hard to extend - if someone wants to add new models, he'd have to think of this place, too. However, this main source is full of suboptimal stuff, so this doesn't matter anyway 😅

dracy21:
Yeah, I know what you mean — it was annoying that you couldn’t understand anything.
You're right about the Type instead of just the mob number. I'll take care of that when I have time.

> Hi, thanks, that's a good idea. However, it has two points for improvement IMHO:
> 
>     1. It's incomplete, not all model names are mapped.
> 
>     2. Instead "UNKNOWN", we could still print out the `Type` value.
> 
> 
> And just a hint... It's hard to extend - if someone wants to add new models, he'd have to think of this place, too. However, this main source is full of suboptimal stuff, so this doesn't matter anyway 😅

Yeah, I know what you mean — it was annoying that you couldn’t understand anything.
You're right about the Type instead of just the mob number. I'll take care of that when I have time.

sven-n:
Btw, I hope you didn't type that all by hand, because with Regex it's done in a minute 😅
Example: https://regex101.com/r/28K3Hc/1
Regex: `(MONSTER_MODEL_)([\w_]*)\s*=\s*\d+.`
Replace by: `case $1$2: return L"$2";`

PR #198, opened 2025-12-08
Refactor angle calculation and pathfinding functions to use modern C++ standard library

mu-nreal:
Replaced manual trigonometry and angle normalization with std::atan2, std::fmod, and helper functions. Introduced NormalizeAngleDegrees, SignedAngleDelta, and StepTowardsAngle to simplify angle operations. Changed sqrtf to std::sqrt, absf to std::fabs, and minf to std::min/std::clamp. Replaced manual mutex lock/unlock with std::lock_guard for RAII-based locking.


sven-n:
Nice, thank you :)

PR #202, opened 2025-12-08
Refactor pointer effect system to use modern C++ random number generation

mu-nreal:
Replaced rand() call with std::mt19937-based RandomInt helper function in anonymous namespace. Changed manual range calculation from rand() % 32 to RandomInt(0, 31) with inclusive min/max parameters for lifetime generation.

PR #222, opened 2025-12-12
Fix RandomInt distribution initialization

mu-nreal:
Refactors RandomInt to create distributions with parameters directly instead of using a static distribution with param_type, fixing incorrect behavior across calls.

Follow-up to #202.


sven-n:
Thanks :)

PR #206, opened 2025-12-09
Add setup to build with CMake and MinGW-w64 (Linux)

mu-nreal:
A CMake setup to cross-compile the Windows client from Linux using a MinGW-w64 toolchain.

**Prerequisites**

  * A working MinGW-w64 toolchain (for example `i686-w64-mingw32-g++`).
  * A MinGW-w64 build of libjpeg-turbo which provides a `libturbojpeg` library (static or import library) on the library search path of your toolchain.
  * Standard Windows / OpenGL libraries shipped with MinGW-w64 (e.g. `opengl32`, `glu32`, `winmm`, `imm32`, `ws2_32`, etc.).

**Example build commands**

From the repository root:

```sh
cmake -S . -B build-mingw \
  -DCMAKE_TOOLCHAIN_FILE=cmake/toolchains/mingw-w64-i686.cmake
cmake --build build-mingw -j$(nproc)
```

If the linker reports `cannot find -lturbojpeg`, install a MinGW-w64 build of libjpeg-turbo (providing `libturbojpeg.a` / `libturbojpeg.dll.a`) or adjust the `target_link_libraries` entry in `Source Main 5.2/CMakeLists.txt` to match the name of the library available on your system.

PR #241, opened 2025-12-13
Refactor Kanturu modules to standardize include casing

mu-nreal:
Standardize include casing.


mu-nreal:
Already addressed in: https://github.com/sven-n/MuMain/pull/206

PR #242, opened 2025-12-14
Refactor CGM_PK_Field to use modern C++ standard library and improve code organization

mu-nreal:
Replaced NULL with nullptr throughout. Replaced switch statement with constexpr std::array<MonsterDefinition, 12> (kMonsterDefinitions) and FindMonsterDefinition helper function. Replaced rand() with std::mt19937 and std::uniform_int_distribution/std::uniform_real_distribution through helper functions (RandomEngine, RandomInt, RandomFloat) in anonymous namespace. Replaced manual include guards with #pragma once.

### DIFF
--- a/Source Main 5.2/source/GM_PK_Field.cpp
+++ b/Source Main 5.2/source/GM_PK_Field.cpp
@@ -55,7 +55,7 @@ namespace
 
     std::mt19937& RandomEngine()
     {
-        static std::mt19937 engine{ std::random_device{}() };
+        thread_local static std::mt19937 engine{ std::random_device{}() };
         return engine;
     }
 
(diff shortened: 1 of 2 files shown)


Source Main 5.2/source/GM_PK_Field.cpp
```diff
@@ -10,8 +17,44 @@
...
+    {
+        int monsterType;
+        EMonsterModelType monsterModelId;
+        int objectModelId;
+        float scale;
+        bool assignLifetime;
+        int lifetime;
+    };
+
+    constexpr std::array<MonsterDefinition, 12> kMonsterDefinitions{ {
```
sven-n, line 34: Nice :)

PR #255, opened 2025-12-24
Right click sell

### LINKED ISSUE #252: Right click to sell items
When using any merchant in game it would be nice to be able to sell items via right click

Mosch0512:
Adds feature to right click sell items in shops.

Open Question should it even be possible to unequip items in shop? possible accidential sell of non valuable items. we could add an additional check if you realy want to sell the item if it was equiped previously and still allow to unequip items on right click when shop is visible. Note this might be a bit more tricky

closes #252


sven-n:
Nice feature, but I have the feeling this is a bit risky. Could we add a condition for a pressed key, e.g. 'Ctrl'?

Mosch0512:
so you mean when holding ctrl and right or left clicken the item in shop only then selling items should be allowed?

i mean i dont the the risky part ... maybe you could explain me in short what you think could go wrong.

we would have one more complicated option of adding this by acutally extending funcionallity of the games ui

sven-n (approved):

sven-n:
Okay, I tested it and it seems like a good improvement. We can only hope that the detection of highly valuable items is always working 😅

PR #257, opened 2025-12-25
Clear guild creation state to prevent data persistence across switches

### LINKED ISSUE #10: Guild Name is not getting reseted
If an other account created an guild on the same client instance the guld info name and image are not reset.

Mosch0512:
- Centralized all guild-related logic by moving the relevant source files into a dedicated guild directory.
- Introduced a shared constants header to replace magic numbers and improve consistency.
- Fixed an issue where guild creation state incorrectly persisted across server switches by refining the guild cache reset logic to trigger explicitly on logout signals.
- Ensured temporary guild creation data is properly cleared on logout.
- Updated the guild creation window to preserve the entered name when navigating between UI states without leaking state across sessions.
- Refactored UI components to use standardized layouts and strongly typed enums for tabs and buttons to improve maintainability.
- Updated all includes and references to match the new guild source file paths.

closes #10


Mosch0512:
@sven-n i updated the description and improved the code to match common sense requirements. Its now ready to be merged :D

PR #262, opened 2026-01-02
Fix 'item option' descriptions for some wings that were not displayin…

DragonicK:
The item options for certain wings were not displayed correctly in the client due to validation being based on the **ExcellentFlags** variable.

image

 In this case, the correct approach is to use the **OptionType** variable instead.

image


sven-n:
Thank you :)

PR #278, opened 2026-01-25
Add MinGW cross-compilation workflow for pull requests

Mosch0512:


.github/workflows/mingw-build-pr.yml
```diff
@@ -0,0 +1,68 @@
...
+      - name: Build
+        run: |
+          cmake --build build-mingw -j$(nproc)
+
+      - name: Upload artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: main-exe-mingw
+          path: build-mingw/src/Main.exe
+          if-no-files-found: error
```
sven-n: Do we really want to upload artifacts for pull requests?

sven-n (changes requested):

.github/workflows/mingw-build-pr.yml
Mosch0512: no i think you are right i removed that part

sven-n (approved):

PR #281, opened 2026-01-28
Sound not working

Mosch0512:
Fixed sound now its working
This branch includes the sound files and the fix


nitoygo:
I think the sound folder is too big to be included in the repo. 
Maybe good to add a separate build step that downloads the files from somewhere.  
And copyright issues may also need to be considered

sven-n:
Please remove the last commit (with force 😅), because that's 200 MB of binaries. Git is very bad at handling this, and once they're in the history, you can't get rid of them anymore.

Mosch0512:
okidoki

sven-n (approved):

Mosch0512:
I think i found some duplicated logic that i have missed let me check that first 🗡️

PR #294, opened 2026-03-18
Fix displaying of system messages by default

olisikh:
System messages (experience, picked up items) are enabled by default after logging in, but no messages are shown.
This PR fixes the issue.


sven-n:
Thanks :)

PR #296, opened 2026-03-18
bug(exp_tracking): fix experience tracking, both in a bar and character screen

olisikh:
This PR fixes experience tracking. 

Fixes a bug when after some level experience starts showing as $current_experience/0 in Character screen. 
Also fixes a bug with the experience progress bar which sometimes stayed at 0 after leveling up.

The problem appeared during casts and reassignment between 32-bit and 64-bit int and floating types, this PR ensures that experience values do not overflow.

### DIFF
--- a/src/source/CharacterManager.cpp
+++ b/src/source/CharacterManager.cpp
@@ -33,6 +33,7 @@ namespace
     constexpr int kSpecialWingType = ITEM_WING + 135;
 
     constexpr int kDefaultClassTextIndex = 2305;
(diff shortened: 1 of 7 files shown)


olisikh:
The last commit fixes an issue where experience bar resets to 0 and stays at zero if character dies.

I have tested it on a couple characters, experience displays correctly for both high level character and low level character, experience bar properly keeps track of experience gained.

olisikh:
Plus another fix for master level character resets, restores the experience bar tracking which was broken.

PR #299, opened 2026-03-22
Bugfix/mu helper

eduardosmaniotto:
Problem: MU Helper was targeting players with skills.
Root Cause: Two locations in WSclient.cpp used IsMonster(c) || IsPlayer(c), adding players to the target list. Target selection functions in MuHelper.cpp had no monster-only filtering.
Also included build client workflow.


.github/workflows/msvc-build.yml
sven-n, line 1: Please revert this file, we already have automated builds

sven-n (changes requested):
Thank you :)

.github/workflows/msvc-build.yml
eduardosmaniotto: I reverted, but this build was different, it was generating the full client to download.

sven-n (approved):

sven-n:
Thanks :)

PR #300, opened 2026-03-22
feat(stats_calc): recalculate character stats based on master skills

olisikh:
- recalculate character stats based on master skills (works after relogin too)
- calculate defense success rate, both pvm and pvp and reflect it on character info pane


sven-n:
Thanks :)
I noticed that a lot of calculation of these values happen in the render methods - this could be optimized in the future 😅

PR #301, opened 2026-03-24
bug(skill_mapping): fix poison scroll tooltip crash by safe book-to-skill mapping

olisikh:
Fixes a bug when pointing cursor over Poison Skill book resulting in consistent client crashes.

This patch introduces strict book-to-skill mapping which removes index-out-of-bounds error possitbility


sven-n (approved):
Thank you :)

Btw, I thought about getting this info from `ITEM_ATTRIBUTE.m_wSkillIndex`, but the problem is, it's `0` for all scrolls and orbs 😅

PR #310, opened 2026-04-01
fix(npc): wandering merchant Zyro proper model + allow interactions

olisikh:
Uses proper model for Wandering Merchant Zyro instead of falling back to a non-interactive Bull model.

### DIFF
--- a/src/source/ZzzCharacter.cpp
+++ b/src/source/ZzzCharacter.cpp
@@ -12810,6 +12810,120 @@ void ReadEquipmentExtended(int Key, BYTE flags, BYTE* Equipment, CHARACTER* pCha
 
 extern int HeroIndex;
 
+namespace
+{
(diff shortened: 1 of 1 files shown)


olisikh:
The Merchant is showing, however he is not interactive, so something else is missing.

jkok25:
'Wandering Merchant Zyro' does not have the same appearance as 'Wandering Merchant Harold'.
'Wandering Merchant Zyro' has the same appearance as 'Moss the Merchant'.
Zrzut ekranu 2026-04-03 154548

olisikh:
Merchant Zyro is now shown with a proper model (same as Merchant Moss in Elbeland), he can be interacted with, but I'm not sure if interactions are as expected.

Well at least he's not showing as a bull anymore :D

Thanks @jkok25 for the hint!

PR #312, opened 2026-04-04
fix: remove FPS-dependent spawn offsets from skill effects

itayalroy:
Some skill effect constructors applied FPS_ANIMATION_FACTOR to spawn offsets instead of only to per-frame movement. At 60 FPS, that made effects start too low or too close to the target, so skills like Meteorite, Cometfall, Decay, etc appeared to fall faster and finish early.


sven-n:
Thanks :)

PR #323, opened 2026-04-12
Fix devil's beam effect

itayalroy:
Devil's beam uses the joint laser effect,
but the joint laser case stopped before the
shared beam setup ran. Removing the break
lets the effect receive the scale, speed,
lifetime, and tail settings that the shared
beam update code expects.

Additionally, remove one extra movement
slowdown from the shared beam path. The effect
already adjusts for higher FPS by increasing
tail count and scaling movement, so the extra
slowdown made the 60 FPS version advance too
little per frame and look blurrier than at
30 FPS. Removing it makes the 60 FPS beam look
identical to 30 FPS.


sven-n:
Thank you :)

PR #326, opened 2026-04-21
Use client-side pathfinding for observers

itayalroy:
Server-side observer path replay is extremely flaky and delayed in practice, while client-side pathfinding is much smoother and more responsive, as shown in the videos below:

Server-side observer pathfinding: https://youtu.be/zo7H19BjdQ0
Client-side observer pathfinding: https://youtu.be/fSTDhl-TtaI

This only changes observer-side visualization. The server remains fully position-authoritative and still validates the active client, so this does not weaken security or makes it any easier to cheat.


sven-n (approved):
Great, thank you! :)

PR #339, opened 2026-04-27
Fix CJK line wrap and strip BOM in SeparateTextIntoLines

Mosch0512:
Reworks #332 (by @Cabbage0211 credit preserved via cherry-pick). His commit fixed Chinese line-wrap; this PR lands it with the review follow-ups, splits the function into a testable unit, and adds a doctest-based PR gate.

## Behavioural fixes
- Replace `(*lpSeek > 255) ? 2 : 1` with `IsFullWidthCharacter(wchar_t)` covering CJK / Hangul / kana / full-width Latin / full-width punctuation. Cyrillic / Greek / Hebrew / Arabic / Latin-Extended now wrap at the full column count instead of half.
- Wrap at the preceding space whenever one was seen on the line, regardless of the overflowing character's width. Mixed strings like `"Hello 你好"` now break cleanly at the space instead of leaving a trailing space when a CJK char triggers the overflow.
- Strip the UTF-8 BOM the cherry-picked commit added to `SceneCommon.cpp` (original file had none).

## Code structure
- Move `SeparateTextIntoLines` and its new helper into `src/source/Text/TextLineWrap.{h,cpp}`. No project dependencies - links cleanly into the test binary.

## Tests
- `tests/` tree using [doctest](https://github.com/doctest/doctest) (single header vendored at `tests/third_party/doctest/`). Gated by `-DBUILD_TESTING=ON`.
- 17 test cases covering: English wrap-at-space, Chinese hard-break, Russian / Greek / Hebrew at correct single-width (regression for the > U+00FF bug), Japanese hiragana, Korean Hangul, full-width Latin, mixed Latin+CJK / Cyrillic+CJK / Latin+Cyrillic, long unbroken word, `iMaxLine` cap, null/zero guards.
- `doctest_discover_tests` registers each `TEST_CASE` as its own CTest entry, so failures show up by name in CI logs.
- `.github/workflows/mingw-build-pr.yml` configures with `BUILD_TESTING=ON` and runs `ctest` through wine - a failing case blocks the PR.
- `/utf-8` is forced on MSVC for test sources so non-ASCII wide-string literals decode as UTF-8 (otherwise MSVC reads them as the system codepage and breaks every non-Latin test).
- `docs/build-guide.md` gains a `Tests` section describing the layout and how to add new modules.

## What did not change
- `SeparateTextIntoLines` signature. All existing callers continue to work via the forward declarations in `SceneCommon.h` / `SceneCore.h`.
- Anything outside the files listed above.


Mosch0512:
@sven-n i have added some tests so that we one when a new branch would breake that :D

PR #352, opened 2026-04-30
Remove rendertexttype

Mosch0512:
None of the keys removed here are read by any code in the client. They're leftover from older formats or an old launcher that was never wired up.

## Dead keys removed

| Key | Why it's dead |
|---|---|
| \`[Graphics] RenderTextType\` | Selected between two text renderers, but \`CUIRenderTextAdvance\` was never written — only a 2021 commented-out \`new CUIRenderTextAdvance\` line. \`CUIRenderText::Create\` has always had one real case. |
| \`[Audio] SoundEnabled\` / \`MusicEnabled\` | Replaced by \`SoundVolume==0\` / \`MusicVolume==0\` (no separate enable flag). |
| \`[Audio] VolumeLevel\` | Legacy single-volume key, predates per-channel \`SoundVolume\` / \`MusicVolume\`. |
| \`[LOGIN] Version\` / \`TestVersion\` | Launcher metadata. The protocol version sent to the server is the 5-byte \`Version\` array hardcoded in \`WSclient.cpp\`; these strings are unrelated. |
| \`[PARTITION]\` (whole section) | Launcher metadata, never read. |

## Code changes
- \`GameConfig\` — drop the \`m_renderTextType\` field, getter, setter, load/save lines, and the \`CfgKeyRenderTextType\` / \`CfgDefaultRenderTextType\` constants.
- \`CUIRenderText::Create\` — drop the dispatch \`int\` param, just construct \`CUIRenderTextOriginal\`. \`m_iRenderTextType\` / \`GetRenderTextType\` removed too.
- \`Winmain.cpp\` — remove \`g_iRenderTextType\` (definition, extern, GameConfig read); call \`Create(g_hDC)\` without the type.
- \`ZzzOpenData.cpp::OpenFont\` — collapse the two-attempt fallback (which only existed to retry with type \`0\`) to one \`Create(g_hDC)\` call.

## Config migration
\`GameConfig::Load\` now calls a small \`RemoveObsoleteKey\` / \`RemoveObsoleteSection\` helper pair (using \`WritePrivateProfileStringW(..., nullptr, ...)\`) to delete each orphaned key/section from existing user config files on next launch. One line per future retired key — no central registry of valid keys to keep in sync.

\`src/bin/config.ini\` is rewritten in the new format so fresh installs are clean from the start.

## What did not change
- \`CUIRenderTextOriginal\` and the \`IUIRenderText\` interface.
- The hardcoded protocol \`Version\` / \`Serial\` byte arrays in \`WSclient.cpp\` (that's the version actually sent to the server; out of scope here).
- Any rendering or audio behaviour — only working code paths are preserved.

### DIFF
--- a/src/source/GameConfig/GameConfig.cpp
+++ b/src/source/GameConfig/GameConfig.cpp
@@ -56,10 +56,16 @@ void GameConfig::Load()
 
     m_zoom = ReadInt(CfgSectionCamera, CfgKeyZoom, CfgDefaultZoom);
 
-    // Strip keys we used to write but no longer use, so user config files
-    // don't accumulate orphans. Append one line per retired key — no central
-    // registry of valid keys to keep in sync.
+    // Strip keys/sections we used to write but no longer use, so user config
+    // files don't accumulate orphans. Append one line per retired key — no
+    // central registry of valid keys to keep in sync.
     RemoveObsoleteKey(CfgSectionGraphics, L"RenderTextType");
+    RemoveObsoleteKey(CfgSectionAudio,    L"SoundEnabled");   // replaced by SoundVolume==0
+    RemoveObsoleteKey(CfgSectionAudio,    L"MusicEnabled");   // replaced by MusicVolume==0
+    RemoveObsoleteKey(CfgSectionAudio,    L"VolumeLevel");    // legacy single-volume key
+    RemoveObsoleteKey(CfgSectionLogin,    L"Version");        // launcher metadata, never read by client
+    RemoveObsoleteKey(CfgSectionLogin,    L"TestVersion");    // launcher metadata, never read by client
+    RemoveObsoleteSection(L"PARTITION");                      // launcher metadata, never read by client
 }
 
 void GameConfig::Save()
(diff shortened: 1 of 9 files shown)


src/source/GameConfig/GameConfig.cpp
```diff
@@ -57,6 +55,17 @@ void GameConfig::Load()
...
+
+    // Strip keys/sections we used to write but no longer use, so user config
+    // files don't accumulate orphans. Append one line per retired key — no
+    // central registry of valid keys to keep in sync.
+    RemoveObsoleteKey(CfgSectionGraphics, L"RenderTextType");
+    RemoveObsoleteKey(CfgSectionAudio,    L"SoundEnabled");   // replaced by SoundVolume==0
+    RemoveObsoleteKey(CfgSectionAudio,    L"MusicEnabled");   // replaced by MusicVolume==0
+    RemoveObsoleteKey(CfgSectionAudio,    L"VolumeLevel");    // legacy single-volume key
+    RemoveObsoleteKey(CfgSectionLogin,    L"Version");        // launcher metadata, never read by client
+    RemoveObsoleteKey(CfgSectionLogin,    L"TestVersion");    // launcher metadata, never read by client
```
sven-n, line 66: Makes sense for us, but how about other MU Clients? They may lose their settings 😉

Mosch0512: other clients save it in registry ;D
and Version and TestVersion is not used at least in our client dont think this would be different for others xD

Mosch0512: most stuff was not even there only Version and TestVersion i think

Mosch0512: what we could do add a MuMainConfig.ini file which is just for our stuff and does not touch the og file at all

sven-n (approved):

PR #362, opened 2026-05-03
Add test to ensure MuEditor is removed when flag is off (#359)

### LINKED ISSUE #359: Add Tests if MuEditor is completly removed when flag is not set
There is a build flag for MuEditor if set MuEditor is available.
The editor should only be available for Server Admins.

We need to add tests to ensure when building without Editor flag that nothing leaked through.

guptapiyush16:
## Summary

This PR adds an automated test to ensure that the `MuEditor` component is completely excluded from the build when the `ENABLE_EDITOR` flag is set to `OFF`. 

It utilizes CMake's `file(GENERATE)` command to statically dump the final list of source files configured for the `Main` target during a build. A lightweight Python script is then hooked into `ctest` to parse those sources and fail the build if any paths containing `MuEditor` accidentally leak through.

## Related issues

Closes #359

## Checklist

- [x] I have built the project locally (see [`docs/build-guide.md`](docs/build-guide.md)).

### DIFF
--- a/src/.gitignore
+++ b/src/.gitignore
@@ -4,7 +4,6 @@
 /*.dat
 *.sdf
 .tmp/
-.tmp\
 *.bak
 *.sbr
 *.tlog
--- a/tests/editor/CMakeLists.txt
+++ b/tests/editor/CMakeLists.txt
@@ -52,4 +52,15 @@ endfunction()
 
 add_subdirectory(text)
 
-add_subdirectory(editor)
+# Add Editor Leak Test (only runs when ENABLE_EDITOR is OFF)
+if(NOT ENABLE_EDITOR)
+    # Generate a text file containing all sources of the Main target
+    file(GENERATE 
+         OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/main_sources.txt" 
+         CONTENT "$<JOIN:$<TARGET_PROPERTY:Main,SOURCES>,\n>")
+
+    # Add a test that runs a Python script to verify MuEditor isn't in those sources
+    find_package(Python3 COMPONENTS Interpreter REQUIRED)
+    add_test(NAME test_editor_leak 
+             COMMAND Python3::Interpreter ${CMAKE_CURRENT_SOURCE_DIR}/editor/test_leak.py "${CMAKE_CURRENT_BINARY_DIR}/main_sources.txt")
+endif()
--- a/tests/editor/CMakeLists.txt
+++ b/tests/editor/CMakeLists.txt
@@ -1,66 +1,16 @@
-# Test tree. New test modules go in their own subdirectory and call
-# add_subdirectory() below. Keep the entry point here minimal — module
-# CMakeLists.txt files own their own sources, link, and call mu_add_test().
-
-include(third_party/doctest/doctest.cmake)
-
-# When cross-compiling Windows binaries on a non-Windows host (CI's Linux
-# runner, WSL/MinGW), wine has to launch the .exe. Setting this once makes
-# both add_test and doctest_discover_tests pick it up via the
-# CROSSCOMPILING_EMULATOR target property.
-if(WIN32 AND NOT CMAKE_HOST_WIN32 AND NOT CMAKE_CROSSCOMPILING_EMULATOR)
-    find_program(WINE_EXECUTABLE wine)
-    if(NOT WINE_EXECUTABLE)
-        message(FATAL_ERROR
-            "BUILD_TESTING=ON on a non-Windows host requires wine to run "
-            "the cross-compiled test binaries. Install it (Ubuntu/Debian: "
-            "sudo apt-get install wine wine32) or configure with "
-            "-DBUILD_TESTING=OFF.")
-    endif()
-    set(CMAKE_CROSSCOMPILING_EMULATOR ${WINE_EXECUTABLE} CACHE STRING "" FORCE)
-endif()
-
-add_library(mu_test_main STATIC main.cpp)
-target_include_directories(mu_test_main PUBLIC third_party/doctest)
-target_compile_features(mu_test_main PUBLIC cxx_std_20)
-# Test sources contain non-ASCII wide string literals. MSVC defaults to the
-# system codepage when reading sources, which mangles UTF-8 characters into
-# wrong codepoints and breaks every non-Latin language test. /utf-8 forces
-# both the source charset and the narrow execution charset to UTF-8; the wide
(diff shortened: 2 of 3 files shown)


Mosch0512:
@guptapiyush16 notify me once all conversations are resolved 👍🏼

Mosch0512 (changes requested):
I like this i will dorp a review soon.

tests/editor/test_leak.py
```diff
@@ -0,0 +1,28 @@
...
+        
+    sources_file = sys.argv[1]
+    
+    with open(sources_file, 'r', encoding='utf-8') as f:
+        sources = f.read().splitlines()
+        
+    # Normalise to forward slashes so the path-segment check is
+    # platform-independent and won't fire just because the repository
+    # happens to be cloned inside a directory called "MuEditor".
+    leaked_files = [s for s in sources if '/MuEditor/' in s.replace('\\', '/')]
```
Mosch0512, line 16: **Test name overstates coverage.**

This catches **file-path leaks** (sources whose path contains `/MuEditor/`). It does NOT catch the other (and arguably more common) way editor code reaches the binary: `#ifdef _EDITOR` blocks inside `src/source/` files. A quick grep finds **59 files** under `src/source/` that gate code on `_EDITOR`. If someone accidentally removes the `if(ENABLE_EDITOR)` guard around the `_EDITOR` compile definition (`src/CMakeLists.txt:217`), every one of those `#ifdef` blocks compiles in and this test still reports PASS.

Two options:
- **Narrow the name to match what it does:** rename the CTest entry to something like `test_editor_dir_not_in_main_sources` and add a one-line docstring saying it only covers path-level leaks.
- **Broaden the test to match its name:** add a second assertion that `_EDITOR` is not in `$<TARGET_PROPERTY:Main,COMPILE_DEFINITIONS>` when `ENABLE_EDITOR=OFF`.

Either is fine - but the current name + scope mismatch will mislead future readers.

tests/editor/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+        OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/main_sources.txt"
+        CONTENT "$<JOIN:$<TARGET_PROPERTY:Main,SOURCES>,\n>")
+
+    # Add a test that runs a Python script to verify MuEditor isn't in those sources
+    find_package(Python3 COMPONENTS Interpreter QUIET)
+    if(Python3_Interpreter_FOUND)
+         add_test(NAME test_editor_leak
+                   COMMAND Python3::Interpreter ${CMAKE_CURRENT_SOURCE_DIR}/test_leak.py "${CMAKE_CURRENT_BINARY_DIR}/main_sources.txt")
+    else()
+         message(STATUS "Skipping test_editor_leak: Python3 interpreter not found")
```
Mosch0512, line 18: **Silent skip defeats the regression guard.**

A guard test that disappears on hosts without Python3 is the worst of both worlds: green CI on those hosts, surprise red on hosts that have it. The whole point of this test is to be unconditional.

I'd suggest dropping the Python dependency entirely and doing it in pure CMake - the logic is one regex check that `cmake -P` handles natively:

```cmake
if(NOT ENABLE_EDITOR)
    file(GENERATE
        OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/main_sources.txt"
        CONTENT "$<JOIN:$<TARGET_PROPERTY:Main,SOURCES>,\n>")

    add_test(NAME test_editor_dir_not_in_main_sources
        COMMAND ${CMAKE_COMMAND}
            -DSOURCES_FILE=${CMAKE_CURRENT_BINARY_DIR}/main_sources.txt
            -P ${CMAKE_CURRENT_SOURCE_DIR}/check_editor_leak.cmake)
endif()
```

with `tests/editor/check_editor_leak.cmake`:

```cmake
file(STRINGS "${SOURCES_FILE}" sources)
set(leaked "")
foreach(s IN LISTS sources)
    string(REPLACE "\\" "/" s_norm "${s}")
    if(s_norm MATCHES "/MuEditor/")
        list(APPEND leaked "${s}")
    endif()
endforeach()
if(leaked)
    message(FATAL_ERROR "MuEditor leaked into Main sources: ${leaked}")
endif()
```

Self-contained, no language hop, no environment gap. If you prefer to keep Python, switch `QUIET` → `REQUIRED` so a missing interpreter fails loudly instead of silently.

src/.gitignore
```diff
@@ -4,7 +4,6 @@
 /*.dat
 *.sdf
 .tmp/
-.tmp\
```
Mosch0512, line 7: **Unrelated change.**

This `.tmp\` deletion isn't related to the editor-leak test. Per `docs/CODING_RULES.md` §10 (keep diffs focused on one concern), please split this into its own commit or its own PR. The change itself is fine - `.tmp\` was never valid gitignore syntax, the `.tmp/` line above already handles the directory - but it doesn't belong here.

tests/editor/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+ # Editor-specific tests only.
+ # Shared test harness setup (doctest include, wine setup, mu_test_main,
```
Mosch0512, line 2: Minor: stray leading space at the start of these two comment lines (` # Editor-specific tests only.`). Should be flush left like the rest of the file.

tests/editor/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+    file(GENERATE
+        OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/main_sources.txt"
+        CONTENT "$<JOIN:$<TARGET_PROPERTY:Main,SOURCES>,\n>")
+
+    # Add a test that runs a Python script to verify MuEditor isn't in those sources
+    find_package(Python3 COMPONENTS Interpreter QUIET)
+    if(Python3_Interpreter_FOUND)
+         add_test(NAME test_editor_leak
+                   COMMAND Python3::Interpreter ${CMAKE_CURRENT_SOURCE_DIR}/test_leak.py "${CMAKE_CURRENT_BINARY_DIR}/main_sources.txt")
+    else()
```
Mosch0512, line 17: Minor: indentation here is 9 spaces vs 4 elsewhere in the file. Consistent 4-space indent inside the `if(Python3_Interpreter_FOUND)` block would match the surrounding style.

Mosch0512 (commented):
Nice looks good :D some stuff to address

Mosch0512:
Replaced by #455

PR #410, opened 2026-05-24
Translations resx

Mosch0512:
## Summary

Adds a single resx-driven translation system. Translations live in `src/Localization/<Group>.<locale>.resx`; the build runs `Tools/ResxGen` to compile them into typed C++ accessors under `namespace I18N::<Group>`. Call sites use those accessors directly. Replaces the legacy editor `i18n::Translator` (JSON), the in-game `GlobalText.bmd`, and the per-language `Dialog_<lang>.bmd` files. Adds a live language dropdown to the options window so the player can switch without restarting; cached widget text refreshes in place.

Ships translations for `en, de, es, id, pl, pt, ru, tl, uk, zh-TW` all languages are just placeholders and translated with bad ai. Only Engish is as it was bevore. The others either did not exist or where lacking (full hand-migration for Editor / Metadata; machine-translation for Game and German Dialog; original-source salvage for Dialog en / pt / es).

### Reading a translated string

```cpp
#include "I18N/All.h"

// Direct accessor (most common):
button->ChangeText(I18N::Editor::SaveSkills);
ImGui::Button(I18N::Game::InventoryIV);
RenderText(0, 0, I18N::Game::Close388, ...);

// Composed via Format() for {0} {1} placeholders:
auto msg = I18N::Format(I18N::Editor::ErrorIndexAlreadyInUse, { idxStr });

// Legacy integer ID (for code that still uses GlobalText[N]-style lookups
// or server-sent indices):
auto text = I18N::Game::Lookup(legacyId);   // const wchar_t*
```

Each entry in resx becomes an `extern const wchar_t*` (or `const char*` for narrow groups). Reading is a pointer dereference - no map lookup, no string compare per call.

### Cached widget labels (auto-refresh on locale switch)

Pass the **address** of the accessor (`&I18N::...`) instead of the accessor itself. The widget caches the current value AND registers as a locale observer; the cache refreshes automatically when the locale changes.

```cpp
m_BtnInventory.ChangeToolTipText(&I18N::Game::InventoryIV, true);
m_BtnSave.ChangeText(&I18N::Editor::SaveSkills);
m_Checkbox.ChangeText(&I18N::Game::AutoHeal);

// Tabbed radio buttons:
std::list<const wchar_t* const*> tabs = {
    &I18N::Game::Hunting, &I18N::Game::Obtaining, &I18N::Game::OtherSettings
};
m_TabBtn.ChangeRadioText(tabs);

// Dialog/dynamic lookup with auto-refresh:
auto slot = I18N::Game::LookupSlot(legacyId);   // const wchar_t* const*
m_AnyButton.ChangeToolTipText(slot, true);
```

Passing a `std::wstring` (the old overload) still works for dynamic strings (character names, computed numbers). It clears any prior slot binding so the observer won't clobber the literal.

### Switching locale

```cpp
I18N::SetLocale("de");                       // -> updates every group + fires observers
const char* current = I18N::GetCurrentLocale();
std::span<const char* const> codes = I18N::GetAvailableLocales();
const char* shown = I18N::GetLanguageDisplayName("de");   // -> "Deutsch"
int idx = I18N::LocaleIndex("de");           // -1 for unknown
```

`GameConfig.UILocale` is the single source of truth; the options window updates it, persists `config.ini`, and calls `SetLocale`.

### Custom observers (for things that aren't widgets)

Use this when you cache text outside the widget system (`wcscpy` into a buffer, a custom-rendered label, etc.).

```cpp
namespace {
void OnLocaleChanged(void* ctx) noexcept {
    auto* self = static_cast<MyThing*>(ctx);
    // re-read I18N::... and refresh whatever cache lives on `self`
}
}

MyThing::MyThing() {
    I18N::RegisterLocaleObserver(&OnLocaleChanged, this);
}
MyThing::~MyThing() {
    I18N::UnregisterLocaleObserver(&OnLocaleChanged, this);
}
```

Observers fire after the group pointers are updated, so reading `I18N::...` inside the callback gives the new locale's value.

### NPC dialog catalogue

NPC dialog text is now a normal I18N group. The structural branching data (number of replies per entry, link / return per reply) is a separate language-agnostic table.

```cpp
#include "I18N/All.h"
#include "GameLogic/Quests/DialogStructure.h"
namespace dlg = GameLogic::Quests::Dialog;

const wchar_t* npcLine = I18N::Dialog::Lookup(idx);
const auto& entry = dlg::GetEntry(idx);
for (int i = 0; i < entry.numAnswer; ++i) {
    const wchar_t* answer = I18N::Dialog::Lookup(dlg::DialogAnswerLegacyId(idx, i));
    int nextIdx = entry.answers[i].link;
    int replyCode = entry.answers[i].returnCode;
    // ...
}
```

## Adding a new string

1. Add `<data name="...">` to **every** `src/Localization/<Group>.<locale>.resx` (default locale `en` is the source of truth; other locales fall back to it if missing).
   ```xml
   <data name="OpenInventory" xml:space="preserve">
     <value>Open inventory</value>
     <comment>legacy_id=421</comment>   <!-- optional, only if call sites use Lookup(int) -->
   </data>
   ```
2. Rebuild. The generator produces `I18N::<Group>::OpenInventory` (the key slugs to a PascalCase identifier).
3. Use it: `m_Btn.ChangeText(&I18N::<Group>::OpenInventory);`.

## Adding a new locale

1. Drop `<Group>.<locale>.resx` next to the others for each group you want covered.
2. Rebuild. The generator picks up the locale, adds it to the master `kLocales` table, and slots it into every group's per-entry value array.
3. The new locale appears in the in-game language dropdown automatically.

## Function reference (`I18N`)

| Function | Purpose |
|---|---|
| `void SetLocale(const char*)` | Switch every group to the given locale; fire observers. Unknown locale falls back to default. |
| `const char* GetCurrentLocale()` | Current locale code; never null. |
| `std::span<const char* const> GetAvailableLocales()` | All locales this build was generated with, default first. |
| `int LocaleIndex(const char*)` | Index of locale in `GetAvailableLocales()`; `-1` if unknown. |
| `const char* GetLanguageDisplayName(const char*)` | Display name in that locale's own language (`"de"` -> `"Deutsch"`). |
| `std::string Format(const char* fmt, {args...})` | Substitute `{0} {1} ...` placeholders. Out-of-bounds indices stay verbatim. |
| `void RegisterLocaleObserver(cb, ctx)` | Register a `void(*)(void*) noexcept` to run after every `SetLocale`. |
| `void UnregisterLocaleObserver(cb, ctx)` | Removes a previously registered (cb, ctx) pair. |
| `I18N::<Group>::<Key>` | Read the current locale's value for `<Key>` (one extern per resx entry). |
| `I18N::<Group>::Lookup(int)` | Legacy-integer-ID lookup; returns `const wchar_t*` / `const char*`. |
| `I18N::<Group>::LookupSlot(int)` | Same as `Lookup` but returns the slot pointer (`const T* const*`) for the widget auto-refresh path. |

## Widget overloads added

| Widget | Slot-pointer overload |
|---|---|
| `CNewUIButton::ChangeText` | `(const wchar_t* const* nameSlot)` |
| `CNewUIButton::ChangeToolTipText` | `(const wchar_t* const* tooltipSlot, bool istoppos = false)` |
| `CNewUIRadioButton::ChangeText` | `(const wchar_t* const* nameSlot)` |
| `CNewUICheckBox::ChangeText` | `(const wchar_t* const* nameSlot)` |
| `CNewUIRadioGroupButton::ChangeRadioText` | `(std::list<const wchar_t* const*>& slotList)` |
| `CNewUIMuHelper::InsertButton` / `InsertCheckBox` | accept slot pointers for label + tooltip |
| `CNewUIHeroPositionInfo::SetButtonInfo` | accept slot pointers for label + tooltip |
| `CNewUIGuardWindow::InitButton` | accepts slot pointer for caption |


Mosch0512:
Pushed `83615662`:

- **Duplicate legacy_id**: `ResxLoader` now throws a clear error at generation time when two keys in the same group carry the same `legacy_id`, naming both keys and explaining why. The build fails before anything is compiled. `CppEmitter` also emits a `constexpr` `static_assert` next to each generated `kLegacyTable` that verifies the IDs are strictly increasing, so a hand-edited generated file out of sync with the resx sources also fails the compile with a message pointing at the affected group.
- **Format brace escape**: `{{` and `}}` collapse to literal `{` and `}` (std::format / Python convention), so translations can carry literal braces alongside placeholders.
- **DialogImporter null guard**: `Path.GetDirectoryName` returning null for a bare filename no longer crashes; the `Directory.CreateDirectory` call is skipped when the parent path is null or empty.

sven-n (approved):

PR #416, opened 2026-05-26
fix(log): write UTF-8 instead of raw wchar_t bytes so MuError.log is readable

andrescolombo:
## Problem

Calls to `WriteFile` in `ErrorReport.cpp` passed `wcslen(...)` directly as the byte count. Since `wcslen` returns a **character** count and `WriteFile`'s `nNumberOfBytesToWrite` expects **bytes** — and `wchar_t` is 2 bytes on Windows — only half of each log entry was written. Lines were silently truncated mid-character, producing corrupted `MuError.log` output.

## Why UTF-8 (not `* sizeof(wchar_t)` and not `CP_ACP`)?

- **Naive fix** (write `wcslen(...) * sizeof(wchar_t)` bytes) produces valid UTF-16 LE, but the file has no BOM. Any plain text viewer treats it as ANSI and renders embedded NUL bytes as spaces — the classic `H e l l o` spaced-out output.
- **`CP_ACP`** works locally but isn't portable. The file's bytes depend on the writer's system codepage (Shift-JIS on a JP machine, Windows-1252 on EN, etc.), so logs collected from different users are ambiguous without knowing each one's locale.
- **`CP_UTF8`** is the same everywhere. ASCII-only logs (the current case) are byte-identical to ASCII. Non-ASCII content (future translated debug text) is encoded portably and renders correctly in Notepad, Notepad++, VS Code, terminals, etc.

## Changes

**`WriteDebugInfoStr`** — `WideCharToMultiByte(CP_UTF8, ...)` into `narrowBuf[MAX_LOG_LINE_BYTES]`, then `WriteFile` the resulting bytes. Refactored to early-exit pattern per coding rule 2. Buffer sized 4096 to handle UTF-8 worst case for a 1024-wchar source (~3 bytes per BMP char, or 4 per surrogate pair).

**`HexWrite`** — same conversion via `narrowLine[MAX_HEX_LINE_BYTES]` for both the mid-loop flush and the final flush. `len` declared once at function scope.

**`CutHead`** — read buffer changed from `wchar_t[128K]` to `char[128K]`. Byte count from `ReadFile` now correctly equals the char count, which also silently fixes two pre-existing latent bugs:
- `lpszBuffer[dwNumber] = '\0'` was using a byte index as a `wchar_t` index
- `dwSize = dwNumber - (lpCut - lpszBuffer)` was subtracting a `char*` pointer difference from a byte count

**`CheckHeadToCut`** — signature and body switched from `wchar_t*` / `wcslen` / `wcschr` / `wcsncmp` to `char*` / `strlen` / `strchr` / `strncmp`. Still correct under UTF-8 because the `###### Log Begin ######` marker is pure ASCII and UTF-8 preserves ASCII bytes verbatim.

## Result

`MuError.log` is now portable UTF-8, readable in any editor on any locale. No BOM handling, no encoding ambiguity, no truncation.

## Review feedback applied (@Mosch0512)

- ✅ `CP_ACP` → `CP_UTF8` (all 3 sites)
- ✅ Named the magic buffer sizes: `MAX_LOG_LINE_BYTES` (4096), `MAX_HEX_LINE_BYTES` (512)
- ✅ Sized `MAX_LOG_LINE_BYTES` to 4096 (was 2048) to safely fit UTF-8 worst case
- ✅ `WriteDebugInfoStr` refactored to early-exit pattern (rule 2)
- ✅ `len` hoisted to function scope in `HexWrite`

---

### DIFF
--- a/src/source/Core/Utilities/Log/ErrorReport.cpp
+++ b/src/source/Core/Utilities/Log/ErrorReport.cpp
@@ -10,6 +10,14 @@
 #include <imagehlp.h>
 #include "ErrorReport.h"
 
+// Max UTF-8 bytes for a single log line. Source buffer is wchar_t[1024]; UTF-8 needs
+// up to 3 bytes per BMP character (and 4 bytes per surrogate pair), so a 1024-wchar
+// input expands to at most ~3072 bytes. 4096 gives comfortable margin.
+constexpr int MAX_LOG_LINE_BYTES = 4096;
+// Hex-dump line buffer. Output is pure ASCII (hex digits, spaces, colons, CRLF),
+// so 1 byte per source wchar is sufficient with margin.
+constexpr int MAX_HEX_LINE_BYTES = 512;
+
 void DeleteSocket();
 
 CErrorReport::CErrorReport()
@@ -50,12 +58,16 @@ void CErrorReport::Destroy(void)
 
 void CErrorReport::CutHead(void)
 {
+    // Log file is UTF-8. The "###### Log Begin ######" marker is pure ASCII, and UTF-8
+    // preserves ASCII bytes verbatim (no multi-byte sequence starts with a byte < 0x80),
+    // so byte-level strchr/strncmp on '#' reliably locates the marker even if other lines
+    // contain multi-byte sequences.
     DWORD dwNumber;
-    wchar_t lpszBuffer[128 * 1024];
-    ReadFile(m_hFile, lpszBuffer, 128 * 1024 - 1, &dwNumber, NULL);
+    char lpszBuffer[128 * 1024];
+    ReadFile(m_hFile, lpszBuffer, sizeof(lpszBuffer) - 1, &dwNumber, NULL);
     //m_iKey = Xor_ConvertBuffer( lpszBuffer, dwNumber);
     lpszBuffer[dwNumber] = '\0';
-    wchar_t* lpCut = CheckHeadToCut(lpszBuffer, dwNumber);
+    char* lpCut = CheckHeadToCut(lpszBuffer, dwNumber);
     if (dwNumber >= 32 * 1024 - 1)
     {
         lpCut = &lpszBuffer[32 * 1024 - 1];
@@ -65,26 +77,26 @@ void CErrorReport::CutHead(void)
         CloseHandle(m_hFile);
         DeleteFile(m_lpszFileName);
         m_hFile = CreateFile(m_lpszFileName, GENERIC_READ | GENERIC_WRITE, FILE_SHARE_READ, NULL, OPEN_ALWAYS, FILE_ATTRIBUTE_NORMAL, NULL);
-        DWORD dwSize = dwNumber - (lpCut - lpszBuffer);
+        DWORD dwSize = dwNumber - static_cast<DWORD>(lpCut - lpszBuffer);
         m_iKey = 0;
         WriteFile(m_hFile, lpCut, dwSize, &dwNumber, NULL);
     }
 }
 
-wchar_t* CErrorReport::CheckHeadToCut( wchar_t* lpszBuffer, DWORD dwNumber)
+char* CErrorReport::CheckHeadToCut(char* lpszBuffer, DWORD dwNumber)
 {
-    const wchar_t* lpszBegin = L"###### Log Begin ######";
-    int iLengthOfBegin = wcslen(lpszBegin);
+    const char* lpszBegin = "###### Log Begin ######";
+    int iLengthOfBegin = static_cast<int>(strlen(lpszBegin));
 
-    wchar_t* lpFoundList[128];
+    char* lpFoundList[128];
     int iFoundCount = 0;
 
-    for (wchar_t* lpFind = lpszBuffer; lpFind && *lpFind; )
+    for (char* lpFind = lpszBuffer; lpFind && *lpFind; )
     {
-        lpFind = wcschr(lpFind, (int)'#');
+        lpFind = strchr(lpFind, '#');
         if (lpFind)
         {
-            if (0 == wcsncmp(lpFind, lpszBegin, iLengthOfBegin))
+            if (0 == strncmp(lpFind, lpszBegin, iLengthOfBegin))
             {
                 lpFoundList[iFoundCount++] = lpFind;
                 lpFind += iLengthOfBegin;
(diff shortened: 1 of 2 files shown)


src/source/Core/Utilities/Log/ErrorReport.cpp
```diff
@@ -113,13 +114,19 @@ void CErrorReport::WriteDebugInfoStr(wchar_t* lpszToWrite)
...
     {
-        DWORD dwNumber;
-        WriteFile(m_hFile, lpszToWrite, wcslen(lpszToWrite), &dwNumber, NULL);
-
-        if (dwNumber == 0)
+        // Convert UTF-16 wide string to ANSI before writing so the log file is plain text,
+        // readable in any editor without encoding configuration. All log messages are ASCII-safe.
+        char narrowBuf[2048];
+        int len = WideCharToMultiByte(CP_ACP, 0, lpszToWrite, -1, narrowBuf, sizeof(narrowBuf), nullptr, nullptr);
+        if (len > 1)  // len includes the null terminator; > 1 means there is actual content
```
Mosch0512: If lpszToWrite exceeds 2048 bytes after conversion, WideCharToMultiByte returns 0 (ERROR_INSUFFICIENT_BUFFER) and the len > 1 guard silently drops the message. The caller Write() uses a 1024-wchar_t scratch buffer so this is fine today, but any future longer format string disappears with no trace. Either size the narrow buffer dynamically off the wide length, or at minimum assert / fall back when conversion fails.

src/source/Core/Utilities/Log/ErrorReport.cpp
```diff
@@ -113,13 +114,19 @@ void CErrorReport::WriteDebugInfoStr(wchar_t* lpszToWrite)
 {
     if (m_hFile != INVALID_HANDLE_VALUE)
     {
-        DWORD dwNumber;
-        WriteFile(m_hFile, lpszToWrite, wcslen(lpszToWrite), &dwNumber, NULL);
-
-        if (dwNumber == 0)
+        // Convert UTF-16 wide string to ANSI before writing so the log file is plain text,
+        // readable in any editor without encoding configuration. All log messages are ASCII-safe.
+        char narrowBuf[2048];
```
Mosch0512: you toched the code. meaning you have to name the magic numbers according to rule 3 in our coding rules. 

I suggest MAX_LOG_LINE_BYTES

src/source/Core/Utilities/Log/ErrorReport.cpp
```diff
@@ -139,14 +146,16 @@ void CErrorReport::HexWrite(void* pBuffer, int iSize)
 {
     DWORD dwWritten = 0;
     wchar_t szLine[256] = { 0, };
+    char narrowLine[512];
```
Mosch0512: same as in line 119

I suggest MAX_HEX_LINE_BYTES

src/source/Core/Utilities/Log/ErrorReport.cpp
```diff
@@ -139,14 +146,16 @@ void CErrorReport::HexWrite(void* pBuffer, int iSize)
...
     int offset = 0;
     offset += mu_swprintf(szLine, L"0x%00000008X : ", (DWORD*)pBuffer);
     for (int i = 0; i < iSize; i++) {
         offset += mu_swprintf(szLine + offset, L"%02X", *((BYTE*)pBuffer + i));
         if (i > 0 && i < iSize - 1) {
             if (i % 16 == 15) {	//. new line
                 offset += mu_swprintf(szLine + offset, L"\r\n");
-                WriteFile(m_hFile, szLine, wcslen(szLine), &dwWritten, NULL);
+                int len = WideCharToMultiByte(CP_ACP, 0, szLine, -1, narrowLine, sizeof(narrowLine), nullptr, nullptr);
```
Mosch0512: len is delcared in line 157 and 168 could be declared in fuction scope not a blocker

src/source/Core/Utilities/Log/ErrorReport.cpp
```diff
@@ -113,13 +114,19 @@ void CErrorReport::WriteDebugInfoStr(wchar_t* lpszToWrite)
 {
     if (m_hFile != INVALID_HANDLE_VALUE)
```
Mosch0512: you cold exit this early would match our coding guide rule 3. not a blocker

Example:
```c++
  void CErrorReport::WriteDebugInfoStr(wchar_t* lpszToWrite)
  {
      if (m_hFile == INVALID_HANDLE_VALUE) return;

      char narrowBuf[kLogLineMaxBytes];
      int len = WideCharToMultiByte(CP_ACP, 0, lpszToWrite, -1, narrowBuf, sizeof(narrowBuf), nullptr, nullptr);
      if (len <= 1) return;

      DWORD dwNumber;
      WriteFile(m_hFile, narrowBuf, len - 1, &dwNumber, NULL);
      if (dwNumber == 0)
      {
          CloseHandle(m_hFile);
          Create(m_lpszFileName);
      }
  }
```

src/source/Core/Utilities/Log/ErrorReport.cpp
```diff
@@ -113,13 +114,19 @@ void CErrorReport::WriteDebugInfoStr(wchar_t* lpszToWrite)
...
     {
-        DWORD dwNumber;
-        WriteFile(m_hFile, lpszToWrite, wcslen(lpszToWrite), &dwNumber, NULL);
-
-        if (dwNumber == 0)
+        // Convert UTF-16 wide string to ANSI before writing so the log file is plain text,
+        // readable in any editor without encoding configuration. All log messages are ASCII-safe.
+        char narrowBuf[2048];
+        int len = WideCharToMultiByte(CP_ACP, 0, lpszToWrite, -1, narrowBuf, sizeof(narrowBuf), nullptr, nullptr);
```
Mosch0512: better to use CP_UTF8 instead of CP_ACP 

With CP_ACP, the meaning of the file's bytes depends on the writer's system codepage. Same code on a JP Windows machine writes Shift-JIS bytes; on EN it writes Windows-1252. A log file you receive from a user is then ambiguous ... you don't know which codepage it's in unless you also know their locale. UTF-8 is the same everywhere.

src/source/Core/Utilities/Log/ErrorReport.cpp
Mosch0512: same as bevore better to use CP_UTF8 instead of CP_ACP

src/source/Core/Utilities/Log/ErrorReport.cpp
```diff
@@ -156,7 +165,8 @@ void CErrorReport::HexWrite(void* pBuffer, int iSize)
         }
     }
     offset += mu_swprintf(szLine + offset, L"\r\n");
-    WriteFile(m_hFile, szLine, wcslen(szLine), &dwWritten, NULL);
+    int len = WideCharToMultiByte(CP_ACP, 0, szLine, -1, narrowLine, sizeof(narrowLine), nullptr, nullptr);
```
Mosch0512: same as bevore better to use CP_UTF8 instead of CP_ACP

Mosch0512 (changes requested):
there is some stuff you should change.
and some optionals based on our coding rules.

Mosch0512 (approved):

Issue #297, opened 2026-03-21
Several bug fixes found while self-hosting: Soul Barrier overflow, AoE PvP target bypass


Picorete:
Hi! I've been self-hosting OpenMU with the open-source MuMain client and found a few bugs while testing PvP and general gameplay. Happy to submit PRs if any of these are useful.

## 1. Soul Barrier / Mana Shield damage overflow
**File:** `AttackableExtensions.cs` line 242

When `SoulBarrierReceiveDecrement` exceeds 1.0, the damage reduction calculation produces negative values that overflow to massive positive numbers (~4 billion damage). Players with Mana Shield active take MORE damage instead of less.

**Fix:** Clamp the reduction value to 0-0.9 and ensure damage never goes below 0.

## 2. Area Skill extraTarget bypasses AreaSkillHitsPlayer check
**File:** `AreaSkillAttackAction.cs` ~line 108

When a player casts an AoE skill (Ice Storm, Decay, etc.) with a selected target, the `extraTargetId` is included in the hit list without checking `AreaSkillHitsPlayer`. This means area skills can damage players even when PvP area damage is disabled in the configuration.

**Fix:** Add `AreaSkillHitsPlayer` validation to the `extraTarget` inclusion check.

Let me know if any of these are worth a PR — I can submit them individually or as a batch. I'm actively working on this codebase and will likely find more issues along the way, so let me know if you'd like me to keep reporting/contributing fixes as I go.

Mosch0512:
@Picorete there have been some prs merged is 2. this resolved now?

ugasricardo:
> @Picorete there have been some prs merged is 2. this resolved now?

No idea, I haven been working on my own fork with a Toncs of custom stuff, if any one needs help let me know, I wanted to help develop new features to this main one but was difficult separate them of my customs one

PR #417, opened 2026-05-26
cherry pick: MuHelper and target fixes 

eduardosmaniotto:
Credit to: https://github.com/andrescolombo/MuMain

## Summary

Several MuHelper improvements: basic attack fallback for non-skill classes, item pickup stuck detection, and monster-blocked pickup handling. Also fixes AoE skill targeting player for Elf (Triple Shot) and Wizard (Ice Storm).
### `ZzzInterface.cpp` — Elf support skill target fix
- **`AttackElf`**: Support skills (HEALING/ATTACK/DEFENSE variants) are now handled in the first switch block, mirroring `AttackWizard`'s `SOUL_BARRIER` pattern. `g_MovementSkill.m_iTarget = SelectedCharacter` is set **unconditionally** — not gated by `CheckAttack()`, which returns `false` for self.
- **`AttackElf`/`AttackWizard`**: The generic fallback at the bottom of both functions now uses `CheckAttack() ? SelectedCharacter : -1` instead of unconditional `SelectedCharacter`, correctly handling non-support skills.
- **`ExecuteSkill`**: Added the missing Elf wall-handling block so `SkillElf` is called when there's a wall obstruction.
- Removed dead code from `AttackElf`'s second switch that always sent `SendRequestMagic(Skill, HeroKey)`.
### `MuHelper.cpp` — New features and fixes
- **Basic attack fallback**: `SimulateBasicAttack` — when no attack skill is available or executable, the helper moves to the target and performs a basic attack. Configurable via `bFallbackBasicAttack`.
- **Skill check gate**: Before calling `SimulateAttack`, the helper now checks `CanExecuteSkill()`.
- **Item pickup stuck prevention**: Increase pickup range from 1.5 to 2.0.

## Related issues

Probably fixes point 2. Area Skill extraTarget bypasses AreaSkillHitsPlayer check in https://github.com/sven-n/MuMain/issues/297

## Checklist

- [x] I have built the project locally (see [`docs/build-guide.md`](docs/build-guide.md)).

### DIFF
--- a/src/source/Engine/Object/ZzzInterface.cpp
+++ b/src/source/Engine/Object/ZzzInterface.cpp
@@ -178,18 +178,6 @@ DWORD g_dwOneToOneTick = 0;
 
 bool g_bGMObservation = false;
 
-static bool IsElfSupportSkill(int skill)
-{
-    return skill == AT_SKILL_HEALING
-        || skill == AT_SKILL_HEALING_STR
-        || skill == AT_SKILL_ATTACK
-        || skill == AT_SKILL_ATTACK_STR
-        || skill == AT_SKILL_ATTACK_MASTERY
-        || skill == AT_SKILL_DEFENSE
-        || skill == AT_SKILL_DEFENSE_STR
-        || skill == AT_SKILL_DEFENSE_MASTERY;
-}
-
 #ifdef LEM_FIX_USER_LOGOUT
 bool g_bExit = false;
 #endif // LEM_FIX_USER_LOGOUT [lem_2010.8.18]
@@ -2500,13 +2488,6 @@ void UseSkillElf(CHARACTER* c, OBJECT* o)
     if (g_MovementSkill.m_iTarget >= 0 && g_MovementSkill.m_iTarget < MAX_CHARACTERS_CLIENT)
     {
         TKey = getTargetCharacterKey(c, g_MovementSkill.m_iTarget);
-    }
-    else if (IsElfSupportSkill(Skill)
-        && SelectedCharacter >= 0
-        && SelectedCharacter < MAX_CHARACTERS_CLIENT
-        && CharactersClient[SelectedCharacter].Object.Kind == KIND_PLAYER)
-    {
-        TKey = CharactersClient[SelectedCharacter].Key;
     }
 
     switch (Skill)
@@ -4898,27 +4879,59 @@ void AttackElf(CHARACTER* c, int Skill, float Distance)
             }
         }
         break;
+
+    case AT_SKILL_HEALING:
+    case AT_SKILL_HEALING_STR:
+    case AT_SKILL_ATTACK:
+    case AT_SKILL_ATTACK_STR:
+    case AT_SKILL_ATTACK_MASTERY:
+    case AT_SKILL_DEFENSE:
+    case AT_SKILL_DEFENSE_STR:
+    case AT_SKILL_DEFENSE_MASTERY:
+        if (SelectedCharacter >= 0 && SelectedCharacter < MAX_CHARACTERS_CLIENT)
+        {
+            if (CharactersClient[SelectedCharacter].Object.Kind != KIND_PLAYER)
+            {
+                Attacking = -1;
+                return;
+            }
+
+            if (c != Hero && !g_pPartyManager->IsPartyMember(SelectedCharacter))
+                return;
+
+            c->TargetCharacter = SelectedCharacter;
+
+            ZeroMemory(&g_MovementSkill, sizeof(g_MovementSkill));
+            g_MovementSkill.m_bMagic = TRUE;
+            g_MovementSkill.m_iSkill = Hero->CurrentSkill;
+            g_MovementSkill.m_iTarget = SelectedCharacter;
+
+            if (!CheckTile(c, o, Distance))
+            {
+                if (PathFinding2(c->PositionX, c->PositionY, TargetX, TargetY, &c->Path, Distance))
+                {
+                    c->Movement = true;
+                    c->MovementType = MOVEMENT_SKILL;
+                    SendMove(c, o);
+                }
+                return;
+            }
+
+            SendRequestMagic(Skill, CharactersClient[g_MovementSkill.m_iTarget].Key);
+        }
+        else
+        {
+            SendRequestMagic(Skill, HeroKey);
+        }
+        SetPlayerMagic(c);
+        return;
     }
     if (SelectedCharacter != -1)
     {
         ZeroMemory(&g_MovementSkill, sizeof(g_MovementSkill));
         g_MovementSkill.m_bMagic = TRUE;
         g_MovementSkill.m_iSkill = Hero->CurrentSkill;
-        if (IsElfSupportSkill(Skill)
-            && SelectedCharacter >= 0
-            && SelectedCharacter < MAX_CHARACTERS_CLIENT
-            && CharactersClient[SelectedCharacter].Object.Kind == KIND_PLAYER)
-        {
-            g_MovementSkill.m_iTarget = SelectedCharacter;
-        }
-        else if (CheckAttack())
-        {
-            g_MovementSkill.m_iTarget = SelectedCharacter;
-        }
-        else
-        {
-            g_MovementSkill.m_iTarget = -1;
-        }
+        g_MovementSkill.m_iTarget = CheckAttack() ? SelectedCharacter : -1;
     }
     if (!CheckTile(c, o, Distance))
     {
@@ -5039,17 +5052,6 @@ void AttackElf(CHARACTER* c, int Skill, float Distance)
         //			UseSkillElf( c, o);
     }
     break;
-    case AT_SKILL_HEALING:
-    case AT_SKILL_HEALING_STR:
-    case AT_SKILL_ATTACK:
-    case AT_SKILL_ATTACK_STR:
-    case AT_SKILL_ATTACK_MASTERY:
-    case AT_SKILL_DEFENSE:
-    case AT_SKILL_DEFENSE_STR:
-    case AT_SKILL_DEFENSE_MASTERY:
-        SendRequestMagic(Skill, HeroKey);
-        SetPlayerMagic(c);
-        return;
     case AT_SKILL_MULTI_SHOT:
     {
         if (!CheckArrow())
@@ -7107,7 +7109,9 @@ int ExecuteSkill(CHARACTER* c, ActionSkillType Skill, float Distance)
         {
             for (int i = EQUIPMENT_WEAPON_RIGHT; i <= EQUIPMENT_WEAPON_LEFT; i++)
             {
-                if (ClassIndex == CLASS_KNIGHT || ClassIndex == CLASS_DARK || ClassIndex == CLASS_DARK_LORD
+                if (ClassIndex == CLASS_KNIGHT
+                    || ClassIndex == CLASS_DARK
+                    || ClassIndex == CLASS_DARK_LORD
                     || ClassIndex == CLASS_RAGEFIGHTER)
                 {
                     bool bOk = false;
@@ -7160,7 +7164,6 @@ int ExecuteSkill(CHARACTER* c, ActionSkillType Skill, float Distance)
                 }
                 if (ClassIndex == CLASS_ELF)
                 {
-                    ZeroMemory(&g_MovementSkill, sizeof(g_MovementSkill));
                     g_MovementSkill.m_bMagic = TRUE;
                     g_MovementSkill.m_iSkill = Hero->CurrentSkill;
                     g_MovementSkill.m_iTarget = CheckAttack() ? SelectedCharacter : -1;
--- a/src/source/Engine/Object/ZzzInterface.cpp
+++ b/src/source/Engine/Object/ZzzInterface.cpp
@@ -178,6 +178,18 @@ DWORD g_dwOneToOneTick = 0;
 
 bool g_bGMObservation = false;
 
+static bool IsElfSupportSkill(int skill)
+{
+    return skill == AT_SKILL_HEALING
+        || skill == AT_SKILL_HEALING_STR
+        || skill == AT_SKILL_ATTACK
+        || skill == AT_SKILL_ATTACK_STR
+        || skill == AT_SKILL_ATTACK_MASTERY
+        || skill == AT_SKILL_DEFENSE
+        || skill == AT_SKILL_DEFENSE_STR
+        || skill == AT_SKILL_DEFENSE_MASTERY;
+}
+
 #ifdef LEM_FIX_USER_LOGOUT
 bool g_bExit = false;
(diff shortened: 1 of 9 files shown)


src/source/MUHelper/MuHelper.cpp
```diff
@@ -854,7 +869,18 @@ namespace MUHelper
...
+        {
+            return 0;
+        }
+
+        const int iSkillIndex = g_pSkillList->GetSkillIndex(iSkill);
+        if (iSkillIndex == -1)
+        {
+            return 0;
+        }
+
```
Mosch0512: **This changes behavior for every skill the helper uses, not just the new basic-attack fallback.**

`SimulateSkill` is called from every skill code path in the helper. Pre-PR it set `m_iSkill = iSkill` (the `ActionSkillType` enum) and ran. Post-PR it bails out with `return 0` if `g_pSkillList == nullptr` OR if `GetSkillIndex(iSkill) == -1`.

The `iSkillIndex` conversion is the right fix - `ZzzInterface.cpp:6955` already does `g_pSkillList->GetSkillIndex(Skill)` for the same purpose, so the helper was sending the wrong value before. But the **silent early-return on `-1`** is new behavior that affects every helper skill, not just basic-attack fallback. If a user has a configured helper skill that the skill list doesn't know yet (e.g. unlearned skill, race condition during skill list init), the helper now does nothing instead of attempting it.

Please confirm:
- Is `GetSkillIndex(iSkill) == -1` a true error state (skill not learned / invalid) that should silently no-op, or should it fall through to the fallback path (which would actually match the PR's stated intent - fall back to basic attack when a configured skill can't be executed)?
- Has this been tested with a typical Elf/DK/Wiz config that worked pre-PR? Easy regression to miss because nothing crashes - the helper just stops casting.

src/source/MUHelper/MuHelper.cpp
```diff
@@ -968,6 +994,82 @@ namespace MUHelper
...
+        if (iWeaponRight >= ITEM_SPEAR && iWeaponRight < ITEM_SPEAR + MAX_ITEM_INDEX)
+        {
+            fRange = BASIC_RANGE_SPEAR;
+        }
+        if (gCharacterManager.GetEquipedBowType() != BOWTYPE_NONE)
+        {
+            fRange = BASIC_RANGE_BOW;
+        }
+
+        SelectedCharacter = iCharIndex;
```
Mosch0512, line 1017: **Cross-layer side effect: MuHelper writes the global `SelectedCharacter`.**

`SelectedCharacter` is the player's UI-level current selection (driven by clicks, target hotkey, etc.). Writing to it from inside the helper's basic-attack simulation means the helper silently changes what the player sees as "selected" - e.g. character info panels, the target box, anything else gated on `SelectedCharacter`.

Per `docs/CODING_RULES.md` §5 (separate concerns by layer), game logic shouldn't be reaching into UI state. If the goal is just to get the basic-attack send-path to use this character, can we route through a non-UI variable, or restore `SelectedCharacter` after the action?

Not necessarily a blocker if this is how the rest of the helper already hijacks targeting (please confirm - a quick grep for `SelectedCharacter =` in `MuHelper.cpp` would tell us), but worth being explicit about.

src/source/UI/NewUI/NewUIMuHelper.cpp
```diff
@@ -1164,13 +1180,16 @@ bool CNewUIMuHelper::Render()
...
-    g_pRenderText->SetFont(g_hFont);
-    g_pRenderText->RenderText(m_Pos.x + 20, m_Pos.y + 347, I18N::Game::UsedExtensionFunction, 0, 0, RT3_SORT_CENTER);
+        g_pRenderText->SetFont(g_hFont);
+        g_pRenderText->RenderText(m_Pos.x + 20, m_Pos.y + 347, I18N::Game::UsedExtensionFunction, 0, 0, RT3_SORT_CENTER);
 
-    g_pRenderText->SetTextColor(0xFF00B4FF);
-    g_pRenderText->RenderText(m_Pos.x + 20, m_Pos.y + 365, I18N::Game::NoExtensionFunctionBeingUsed, 0, 0, RT3_SORT_CENTER);
+        g_pRenderText->SetTextColor(0xFF00B4FF);
+        g_pRenderText->RenderText(m_Pos.x + 20, m_Pos.y + 365, I18N::Game::NoExtensionFunctionBeingUsed, 0, 0, RT3_SORT_CENTER);
+    }
```
Mosch0512, line 1192: **UI scope creep: this hides the "Used Extension Function" panel on tab 0.**

The `if (m_iCurrentOpenTab != 0)` guard is a behavior change unrelated to the basic-attack-fallback feature: it removes a panel that previously rendered on every tab. Plus the buff-icon Y coordinates and `RenderBack` rectangles change by 10-15px on lines 326-328 and 1220-1221, presumably to make room for the new checkbox.

Per `docs/CODING_RULES.md` §10 ("Keep changes focused on one concern per commit"), this should either:
- Be called out explicitly in the PR description ("layout shifts to accommodate new checkbox; the extension-function panel is now hidden on tab 0"), so reviewers know it's deliberate, or
- Be split into a separate UI-layout commit/PR.

The checkbox-fitting layout shifts are unavoidable; the `if (m_iCurrentOpenTab != 0)` panel hide looks separate. What's the reason for it?

src/source/Network/Server/WSclient.h
```diff
@@ -3645,7 +3645,14 @@ typedef struct
...
     BYTE PetAttack;                          // Index: 32
-    BYTE _UnusedPadding[36];
+    
+    BYTE bUseSelfDefense : 1;                // Index: 33 (bit 0)
+    BYTE bAutoAcceptFriend : 1;              // Index: 33 (bit 1)
+    BYTE bAutoAcceptGuild : 1;               // Index: 33 (bit 2)
+    BYTE bFallbackBasicAttack : 1;           // Index: 33 (bit 3)
+    BYTE : 4;                                // Unused bits of Index 33
+
+    BYTE _UnusedPadding[35];                 // Index: 34 (35 bytes remaining)
```
Mosch0512, line 3655: **Protocol bits land in what was server padding - verified wire-compatible.**

I checked OpenMU: `MuHelperConfiguration` is stored as `byte[]` on `Character` and echoed back unchanged on login (`Player.cs:2516`). The server doesn't parse byte index 33, but it also doesn't strip it - so the four new bits persist for the same client across reconnects.

No server change is required **provided all four settings are intended to be client-local** (the field names suggest yes - auto-accept friend/guild, self-defense, fallback all sound like client behaviors). If any of them are meant to be server-enforced (e.g. server auto-accepting friend requests on the player's behalf when offline), then a coordinated OpenMU change is needed too.

Please confirm which of these four are client-local-only vs. server-enforced. As-is, the comment block reads as if a server change is expected - clarifying the intent here would save a future maintainer from chasing the server side.

Mosch0512 (dismissed):

src/source/Engine/Object/ZzzInterface.cpp
```diff
@@ -7135,6 +7160,10 @@ int ExecuteSkill(CHARACTER* c, ActionSkillType Skill, float Distance)
                 }
                 if (ClassIndex == CLASS_ELF)
                 {
+                    ZeroMemory(&g_MovementSkill, sizeof(g_MovementSkill));
+                    g_MovementSkill.m_bMagic = TRUE;
+                    g_MovementSkill.m_iSkill = Hero->CurrentSkill;
+                    g_MovementSkill.m_iTarget = CheckAttack() ? SelectedCharacter : -1;
```
Mosch0512: **New `g_MovementSkill` initialization for the ELF block.**

Pre-PR, the `if (ClassIndex == CLASS_ELF)` block at this position called `SkillElf` directly without touching `g_MovementSkill`. The CLASS_DARK_KNIGHT block above did initialize it (for `SkillWarrior`); the ELF block didn't.

Post-PR, the ELF block does the same init. This is a behavior change inside the cherry-pick that isn't covered by the PR description. Two possibilities:

1. **Bug fix:** `SkillElf` was relying on stale `g_MovementSkill` state, and this initializes it correctly. So great, but please mention it in the PR body so it's not invisible.
2. **Cherry-pick collateral:** the upstream change refactored target handling broadly and this got dragged along. Worth a check that `SkillElf` doesn't depend on the *prior* contents of `g_MovementSkill` (e.g. from an earlier code path that already set it for the same tick).

A quick read of `SkillElf` (`ZzzInterface.cpp:3173-3210+`) didn't show obvious dependence on `g_MovementSkill`, but I'd appreciate the author confirming this was tested with each Elf skill the helper triggers (heal, buff, attack).

src/source/Engine/Object/ZzzInterface.cpp
```diff
@@ -4885,7 +4904,21 @@ void AttackElf(CHARACTER* c, int Skill, float Distance)
...
+        {
+            g_MovementSkill.m_iTarget = SelectedCharacter;
+        }
+        else if (CheckAttack())
+        {
+            g_MovementSkill.m_iTarget = SelectedCharacter;
+        }
+        else
+        {
+            g_MovementSkill.m_iTarget = -1;
```
Mosch0512: **Target-decision logic is now duplicated across four sites.**

The same `if (IsElfSupportSkill && in-range && KIND_PLAYER) then SelectedCharacter / else if (CheckAttack) then SelectedCharacter / else -1` pattern (or a subset of it) lives here, at `AttackWizard:6298`, and in `ExecuteSkill` for both `CLASS_DARK_KNIGHT` and `CLASS_ELF` (lines 7154, 7166). Per `docs/CODING_RULES.md` §4 (Write Reusable Code), extract:

```cpp
int DecideMovementTarget(int skill, bool allowSupportTargeting)
{
    if (allowSupportTargeting && IsElfSupportSkill(skill)
        && SelectedCharacter >= 0 && SelectedCharacter < MAX_CHARACTERS_CLIENT
        && CharactersClient[SelectedCharacter].Object.Kind == KIND_PLAYER)
    {
        return SelectedCharacter;
    }
    return CheckAttack() ? SelectedCharacter : -1;
}
```

Then each site collapses to `g_MovementSkill.m_iTarget = DecideMovementTarget(Skill, /*allowSupport=*/ClassIndex == CLASS_ELF);` and any future tweak to the rule lives in one place instead of four. This isn't blocking but the four-site duplication makes future drift very likely.

Mosch0512 (commented):
Re-posting three inline comments that were collapsed/dismissed by accident. Content identical to the original review, with the 6298 comment in its updated form (the buff-party-member regression).

Mosch0512:
@eduardosmaniotto  sry i am using this pr/review for testing out claude for reviewing but i had some problems as you might have seen ^^

src/source/Network/Server/WSclient.h
eduardosmaniotto: I can confirm that the server doesn't enforce this packet. It stores everything as is. Adding new bits will not break anything, but changing previously created ones will break offline leveling since it uses the Mu Helper configuration for the server-side bot behavior. Comments added for clarification.

src/source/UI/NewUI/NewUIMuHelper.cpp
eduardosmaniotto: It's hiding because the new checkbox doesn't fit on the window. In fact, I think it could be completely removed since that section is currently doing nothing. It's related to the attached image block:
image

src/source/MUHelper/MuHelper.cpp
Mosch0512: most likely false positive after testing

src/source/UI/NewUI/NewUIMuHelper.cpp
Mosch0512: ok thats fine than :D

src/source/MUHelper/MuHelper.cpp
Mosch0512: also fasle postiv

src/source/Engine/Object/ZzzInterface.cpp
eduardosmaniotto: I completely removed the IsElfSupportSkill because I believe it was an unnecessary workaround for something that should be handled by the AttackElf (which already had the switch statements for buff skills). Now it mirror the AttackWizard behavior.

src/source/Engine/Object/ZzzInterface.cpp
eduardosmaniotto: I tested without the code and I had problems with the triple shot AoE hitting other players unintentionally. I'm not 100% sure, since I couldn't reliably replicate the bug all the time, but at least with this change it never happened. So I believe it's a valid bug fix.

src/source/MUHelper/MuHelper.cpp
```diff
@@ -968,6 +983,82 @@ namespace MUHelper
...
+            return 0;
+        }
+
+        const bool bTargetNear = CheckTile(Hero, &Hero->Object, fRange);
+        const bool bNoWall = CheckWall(Hero->PositionX, Hero->PositionY, TargetX, TargetY);
+
+        if (!bTargetNear || !bNoWall)
+        {
+            Hero->Path.Lock.lock();
+            const int pathNum = std::min<int>(tempPath.PathNum, 2);
```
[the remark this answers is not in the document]
Mosch0512: we will merge this part also even though i am not sure about what happens when we have players and one is moving manauly vs helper this sounds like helper has an unfair advantage but i think it should be the opposite around.

player can press and hold alt to click through monsters and pick up items not clickable otherwise.

@eduardosmaniotto what do you think

Mosch0512 (commented):
I noticed especcialy when BAsic attack fallback is used that user cant manualy move anymore i think we should always prio manuell movements meaning when player trys to move he can move. right now its basically prevented when using the bot ... 

one idea would be to have a cooldown so that if player moves manually or doth avitvate a skill manually or anything there is a pause for the bot for 5sec which gets set to 5sec each time player does sth this would make the bot much more responsive if you need to take action quickly

but i think this is to big for this branch maybe we do this in a new issue.

src/source/MUHelper/MuHelper.cpp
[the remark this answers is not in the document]
eduardosmaniotto: I'll reply to this one since the other comment doesn't give the option.
Regarding the item pick change, I reverted it because the more I looked at it, the more it felt like a hack instead of a proper fix. So I changed the range instead and found no issues while testing. In relation to the advantage, manual players can press the spacebar, which gets items even faster than Alt+Click and further than the bot (Space picks up from 3 tiles). Also, when I played high-ping servers (playing a European server from America), I had no chance picking up items against players with lower pings, so the ping is still the biggest factor in this.

In relation to the manual movement while the bot is running, I agree that it's currently janky, but from my experience it's even worse on some private servers, where you can't even open the inventory with the bot running. I also believe this would be a bigger change; I'm not sure how much, so it would be a good idea to separate it in another PR.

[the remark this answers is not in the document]
Mosch0512: i love you

PR #432, opened 2026-06-02
Bugfix: fallback atk speed

### LINKED ISSUE #430: Basic Attack Fallback Mi helper hitting to fast
I am not sure If I had a custom attack speed on the weapon but it looked like the bot can hit basic attacks way faster than he should be able to.

eduardosmaniotto:
## Summary

SimulateBasicAttack is sending SendHitRequest on every helper loop tick when the target was in range, ignoring whether the character's attack animation was still playing. This caused the fallback to hit ~2–3× faster than a normal manual basic attack, since the manual path is naturally rate-limited by the animation frame counter.

## Related issues

Closes #430 

## Checklist

- [X] I have built the project locally (see [`docs/build-guide.md`](docs/build-guide.md)).

### DIFF
--- a/src/source/MUHelper/MuHelper.cpp
+++ b/src/source/MUHelper/MuHelper.cpp
@@ -955,6 +955,11 @@ namespace MUHelper
             return 0;
         }
 
+        if (Hero->AttackTime > 0.0f)
+        {
+            return 0;
+        }
+
         const int iCharIndex = FindCharacterIndex(iTarget);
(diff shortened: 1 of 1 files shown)


Mosch0512:
@eduardosmaniotto i have tested it sadly the attack speed is still about 2x faster than normal attacks manaualy done tested with new char

Mosch0512:
i will fix this @eduardosmaniotto

PR #437, opened 2026-06-05
Refactor zzzinterface decomposition

Mosch0512:
some more cleanup


Mosch0512:
all of the reviews are exiting errors not introduced with this refactor i will skip them

Issue #396, opened 2026-05-17
Bugged shadows


itayalroy:
Player shadows look completely messed up, not sure if this is caused by camera refactor or not.

<img width="541" height="494" alt="Image" src="https://github.com/user-attachments/assets/6a776a12-164b-44f6-9bc0-c0c1da7f2e53" />

Mosch0512:
i just checked if its old or new and the bug was there before camera 3d so its old

Mosch0512:
Closed by #448

PR #448, opened 2026-06-06
Fix equipped weapon shadows

itayalroy:
Render weapon shadows through the linked item path instead of the body-part path.

Weapons are separate item models attached to the character with RenderLinkObject. The old shadow pass used RenderPartObject, which is meant for body parts, so weapon shadows used the wrong position and rotation math and appeared detached from the player.

This keeps detailed weapon shadows, but makes them follow the same attachment logic as the visible weapon.

Fixes issue https://github.com/sven-n/MuMain/issues/396

image


Mosch0512:
@itayalroy does this also fix cape shadow ? i will look at it asap

itayalroy:
@Mosch0512 this pr does not affect cape shadows, it only touches weapons

Mosch0512:
okay i will create an seperate issue for that  cape shadow than

PR #457, opened 2026-06-07
Add portable single-line text field and use it for login

### LINKED ISSUE #447: Replace legacy Win32 EDIT-control text input with a portable text-input system
## Goal
Replace the legacy Win32 child `EDIT` controls used for all in-client text entry with a portable text-input system (SDL3 text input + custom-rendered input boxes), so text entry works off Windows and the SDL windowing path (#442) can drop its transitional Win32 message hook.

## Current state
`CUITextInputBox` / `CUIEditControl` create real Win32 child windows:

- `src/source/UI/Legacy/UIControls.cpp:3386` - `CreateWindowW(L"edit", … WS_CHILD, m_hParentWnd, …)`, subclassed via `EditWndProc` (UIControls.cpp:3108), with Win32 IME (`ImmGetContext(m_hEditWnd)`, composition-window positioning).
- Used for every text field: login username/password (`LoginWin`), chat input (`NewUIChatInputBox`), shop title, MU Helper inputs, guild/friend windows, etc.

These controls require the Win32 message pump (`TranslateMessage` / `DispatchMessage`) and a parent `WndProc` to receive keystrokes and IME composition. They do not exist off Windows, and they are the main reason #442's SDL windowing PR keeps a Windows-only `SDL_SetWindowsMessageHook` bridge.

## Proposed change
- Implement a portable text-input field: capture text via SDL `SDL_StartTextInput` / `SDL_EVENT_TEXT_INPUT` / `SDL_EVENT_TEXT_EDITING`, maintain the edit buffer / caret / selection in C++, and render it with the existing UI text renderer (no Win32 child window).
- Replace `CUITextInputBox` / `CUIEditControl` usages with the portable field.
- Route IME through SDL text editing (composition string) instead of `Imm*` on `m_hEditWnd`.
- Render the selection with explicit, readable colors. The plain Win32 `EDIT` draws selected text with the OS `COLOR_HIGHLIGHT` / `COLOR_HIGHLIGHTTEXT`, which are near-illegible against the dark (black background / white text) boxes and cannot be overridden per-control; the portable field should pick its own selection background/text colors.

## Why / sequencing
- Depends on #442 (SDL windowing). Removing the legacy controls is what lets #442's Windows message hook be deleted and enables text entry on Linux/macOS.
- Large and cross-cutting (touches every text field) - comparable in size to the `wchar_t` width effort.

## Out of scope
- The `wchar_t` width difference and other non-window Win32 APIs (registry, `wininet`/`urlmon`, DPAPI) - separate issues.

## Acceptance
- All text fields (login, chat, shop, MU Helper, guild/friend) work with no Win32 child controls.
- Windows behaves as before (including IME); a Linux build accepts text input.
- Selected text is readable: the input boxes use their own selection background/text colors instead of the OS highlight colors.
- #442's `SDL_SetWindowsMessageHook` input/IME bridge can be removed.

Mosch0512:
Introduce a self-rendered, SDL-fed text field in CUITextInputBox that runs alongside the legacy Win32 EDIT control and is selected per instance with EnablePortableInput(). The login username and password boxes opt in; every other consumer keeps using the EDIT path.

The portable field owns its text buffer, caret, selection and horizontal scroll, renders through g_pRenderText with manually drawn selection and caret rectangles, and keeps password masking plus the NUMBERONLY, SERIALNUMBER and NOLOCALIZEDCHARACTERS filtering and full-to-half-width conversion. Every shared method early-branches on the portable flag so the EDIT path is unchanged for boxes that have not opted in.

The SDL event loop routes SDL_EVENT_TEXT_INPUT and SDL_EVENT_KEY_DOWN to the focused field, handles Ctrl+A/C/V/X through the SDL clipboard so the field stays free of SDL, and starts or stops SDL text input as a portable field gains or loses focus. The login window drops focus when hidden so a hidden screen no longer holds keyboard input.

Closes #447

### DIFF
--- a/src/source/UI/NewUI/NewUIManager.cpp
+++ b/src/source/UI/NewUI/NewUIManager.cpp
@@ -1,5 +1,6 @@
 #include "stdafx.h"
 #include "UI/NewUI/NewUIManager.h"
+#include "UI/Legacy/UIControls.h"  // CUITextInputBox::GetFocusedPortable (issue #447)
 
 
 using namespace SEASON3B;
@@ -147,6 +148,14 @@ bool SEASON3B::CNewUIManager::UpdateKeyEvent()
 
     auto vecUI = m_vecUI;
 
+    // Portable text fields (issue #447) don't take Win32 focus, so GetFocus()
+    // can't tell which widget owns the keyboard. When a field is focused, use
+    // its stable identity (matches the owning widget's GetRelatedWnd(), which it
+    // set to the same handle); otherwise fall back to the window focus. This
+    // keeps game hotkeys from firing while the player is typing.
+    CUITextInputBox* pFocusedField = CUITextInputBox::GetFocusedPortable();
+    const HWND hFocus = pFocusedField ? reinterpret_cast<HWND>(pFocusedField) : GetFocus();
+
     auto vi = vecUI.begin();
     for (; vi != vecUI.end(); vi++)
     {
@@ -156,7 +165,7 @@ bool SEASON3B::CNewUIManager::UpdateKeyEvent()
             hRelatedWnd = g_hWnd;
         }
 
-        HWND hWnd = GetFocus();
+        HWND hWnd = hFocus;
 
         if ((*vi)->IsEnabled() && hWnd == hRelatedWnd)
         {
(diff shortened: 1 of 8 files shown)


src/source/UI/NewUI/NewUIManager.cpp
```diff
@@ -287,7 +296,7 @@ bool SEASON3B::CNewUIManager::CompareKeyEventOrder(INewUIBase* pObj1, INewUIBase
 int SEASON3B::CNewUIManager::GetShowUICnt()
 {
     int m_nShowUICnt = 0;
-    // �Ϻ� Ư�� �������̽��� � �����ִ���
+    // �Ϻ� Ư�� �������̽��� � �����ִ���
```
Mosch0512: How many of certain interfaces are open

in english

PR #493, opened 2026-06-10
Add data-driven effect registry and route effect dispatch through it

sven-n:
Introduce EffectDescriptor/EffectRegistry: a per-type table of creation
parameters plus optional move/render handlers. CreateEffect, MoveEffect and
RenderEffects now consult the registry first and fall back to the legacy
switch statements for types that haven't been migrated yet, so effects can be
converted from hard-coded cases to data one at a time.

Migrate MODEL_DESAIR end-to-end as the first effect to exercise all three
stages (create params, move handler, default render).


src/source/Render/Effects/Behaviors/MoveHandlers.cpp
```diff
@@ -0,0 +1,10506 @@
...
+
+    // BITMAP_OUR_INFLUENCE_GROUND, BITMAP_ENEMY_INFLUENCE_GROUND
+    bool Move_BITMAP_OUR_INFLUENCE_GROUND(OBJECT* o, int index, float Luminosity)
+    {
+        if (o->SubType == 0)
+        {
+            if (o->Owner == NULL || o->Owner->Live == false)
+                o->Live = false;
+
+            VectorCopy(o->Owner->Position, o->Position);
```
Mosch0512, line 1464: Noticed this while reading through: if `Owner` is NULL we set `Live = false`, but the very next line still dereferences `o->Owner->Position`. It's a faithful copy of the old switch (same latent bug there), but since the other handlers got owner guards, maybe worth an early `return true;` here too? ^^

Issue #462, opened 2026-06-07
Native Linux build: port the engine off Win32 (wide-char + platform APIs)


Mosch0512:
## Goal
Make the client actually build and run natively on Linux, not just the portable SDL shell. This is the large follow-up that #442 deferred: #442 routed windowing/input/GL/event-loop/text/IME through SDL (the portable shell), but the rest of the engine is still Windows-bound.

## Current state
- #442 made the shell portable. `Platform::SdlShell` + the `mu_sdl_smoke` target (window + GL + event loop, SDL3 only, no engine) build and run.
- **Phase 0 done (#463, merged):** native Linux configure + build + run of `mu_sdl_smoke` works with no code changes (the static-turbojpeg `FATAL` is `WIN32`-gated; Python + .NET cover the wire-size codegen). A `linux-smoke` CI job guards it.

## Triage (native Linux compile of the engine, `MuClient` target)
The engine PCH `stdafx.h` already guards `windows.h` behind `#ifdef _WIN32` (prior groundwork). Remaining categories, by file count:

| Category | Scope | Notes |
|---|---|---|
| `#include <windows.h>` (unguarded) | **21 headers -> 352 files fail transitively** | Widely-included headers (`Scenes/SceneCommon.h`, `Core/Utilities/StringUtils.h`, `Data/GameData/.../*Structs.h`, ...). The shallow include wall hiding everything below. |
| `wchar_t` is 16-bit (Win) vs 32-bit (Linux) | pervasive | **The big one.** Affects packet/wire structs, string buffers, file/clipboard I/O. |
| `MessageBox` | 149 files | Error/info dialogs. Route through a portable box (e.g. `SDL_ShowSimpleMessageBox`). |
| winsock (`SOCKET`, `WSAStartup`, ...) | 23 files | Networking; needs a BSD-socket abstraction. |
| winmm timers (`timeGetTime`/`timeBeginPeriod`) | 22 files | Portable via `SDL_GetTicks` / `std::chrono`. |
| MSVC secure-CRT (`sprintf_s`, `wcscpy_s`, ...) | 18 files | Provide portable equivalents or migrate to standard calls. |
| `wininet`/`urlmon` (in-game shop downloader) | 7 files | Replace with libcurl (or disable the downloader on Linux). |
| Win32 file API (`CreateFileW`/`ReadFile`/...) | 7 files | Portable via `std::fstream` / POSIX. |
| `CreateThread` | 2 files | -> `std::thread`. |
[...]

PR #495, opened 2026-06-13
Run and connect the client natively on Linux (#462)

Mosch0512:
Brings the Linux port to a running, connecting client (issue #462, Phase 4). The engine now boots into the login scene, renders through SDL/OpenGL, reads its config, and connects to a server on Linux. Windows builds are unchanged: every change is either guarded by `#ifdef _WIN32` / `#ifndef _WIN32` or is a platform-neutral correctness fix.

## .NET client library on Linux
- Build `MUnique.Client.Library` as a native `linux-x64` Native AOT shared object and `dlopen` the `.so` instead of the Windows `.dll`.
- Fix native-to-managed string marshalling. The native client passes `wchar_t*`, which is 4-byte UTF-32 on Linux, but the library decoded it as UTF-16. This garbled the server address into a bogus IP, so `connect()` hung forever on the main thread (which looked like a black screen). Strings are now decoded by the platform's real `wchar_t` width. The XSLT template is updated so regeneration stays correct.

## Runtime fixes uncovered by running
- Config was never read on Linux: the executable-directory split looked for a backslash only, so `config.ini` was never found and every setting silently fell back to its default. It now accepts either path separator.
- UI piled into the top-left corner: `CInput::Create` bailed on the null SDL window handle before setting the screen size, so centered windows computed negative positions. The size is now set regardless of the handle off Windows.
- LP64 width bugs reading binary assets (on-disk size fields are 32-bit, but `long`/`size_t` are 8 bytes on Linux x64): pet.bmd speed array, BMD encrypted size, and the camera walk-script waypoint count. These were latent on any 64-bit build.
- Double `fclose()` of the same handle (filter file, object files): tolerated on the old CRT, a double-free on glibc.
- `mu_swprintf` is bounded by the real array size to stop a fortify abort, and a frame can be dumped via `MU_CAPTURE_FRAME` for headless verification.

## Portable subsystems
- GDI text backend backed by stb_truetype (`CreateFont`/`CreateDIBSection`/`TextOut`) so UI text renders without Win32 GDI.
- Case-correcting path resolver so Windows-style asset paths resolve on a case-sensitive filesystem, wired through every file-open shim.
- GameShop compiles on Linux: the UI and list managers build, only the WinINet/urlmon downloader is excluded, with the download entry points failing gracefully.
- Extended portable shims (StrSafe, Sleep/GetLastError, file-attribute and directory helpers, crash-log system info), and the Linux entry point now calls the shared `WinMain` bootstrap.

## Known follow-ups
- The CMake `ClientLibrary` target still targets the Windows DLL via `dotnet.exe`; wiring it to build the `linux-x64` `.so` natively when running on Linux is the next step so a plain `cmake --build` produces a working client.


Mosch0512 (commented):
## Overview

Brings the Linux port to a running, connecting client: builds the `.NET` client library as a `linux-x64` Native AOT `.so`, fixes the `wchar_t` (UTF-32 on Linux) marshalling that garbled the server address, adds portable backends for GDI text (stb_truetype), case-correcting path resolution, and a batch of LP64 width / double-`fclose` correctness fixes. The Win32 paths are mostly fenced behind `#ifdef _WIN32`. It's a large but well-commented and coherent change, and the structure (one shim header per Win32 subsystem) is clean.

I verified the headline correctness fixes:
- The two removed `fclose(fp)` calls (`ZzzInfomation.cpp`, `ZzzObject.cpp`) are genuine double-closes - in both functions `fp` is already closed right after the read, so this is correct.
- The LP64 width fixes (pet speed array, BMD enc size, camera waypoint count) all match a 32-bit on-disk field, correct.
- The UTF-32 `wchar_t` marshalling fix is sound and the XSLT template is updated so regen stays correct.

A few things below are worth a look - the first is the most important because it appears to change Windows behavior, which the PR description says it doesn't.

## Findings

See inline comments. Summary:

1. **F10 (and Enter) added to the cross-platform `MainLoop` are unguarded** - on Windows these likely fire in addition to `WndProc`, double-toggling the zoom lock. (correctness, Windows regression)
2. **`dlopen("./MUnique.Client.Library.so")`** resolves against the CWD, not the binary directory as the comment claims. (robustness)
3. **`GetCurrentThreadId` truncates `pthread_self` to 32 bits.** (minor)
4. **CI workflows trigger on both `push` and `pull_request`** with no branch filter - same-repo PR branches get every job run twice. (CI cost)

Nits (not blocking): `.gitignore` now lists both `_deps/` and `_deps`; `MoveFileExW` ignores its flags and uses `rename` unconditionally (fine for the same-dir backup callers, but cross-filesystem moves would `EXDEV`).

src/source/App/Platform/Windows/Winmain.cpp
```diff
@@ -1041,6 +1099,25 @@ MSG MainLoop()
...
+                    SetEnterPressed(true);
+                }
+                // F10 toggles the camera zoom lock. On Windows this is handled
+                // in WndProc's WM_SYSKEYDOWN (F10 is a reserved system key); the
+                // SDL path does not drive WndProc, so without this the zoom stays
+                // locked and the mouse wheel can never zoom. Edge-triggered, like
+                // the WM_SYSKEYDOWN PREVIOUS_KEY_STATE check.
+                if (event.key.scancode == SDL_SCANCODE_F10 && !event.key.repeat)
+                {
+                    CameraManager::Instance().ToggleZoomLock();
```
Mosch0512, line 1122: This F10 toggle (and the `SetEnterPressed` block just above) is added to `MainLoop`, which runs on Windows too - it is not under `#ifndef _WIN32`. On Windows `WndProc` still handles F10 via `WM_SYSKEYDOWN` (line ~522) and `WndProc` is driven from SDL through `SDL_SetWindowsMessageHook(Win32MessageHook, ...)`. So a single F10 press looks like it would reach **both** handlers and call `ToggleZoomLock()` twice -> net no-op, i.e. F10 stops working on Windows.

The inline comment ("the SDL path does not drive WndProc") is true on Linux but not on Windows, where the message hook is installed. The Enter case is harmless because `SetEnterPressed(true)` is idempotent, but the F10 toggle is not.

The PR description states "Windows builds are unchanged". Could you confirm on Windows whether F10 double-fires? If so, guard this block with `#ifndef _WIN32` (Enter could stay either way).

src/source/Dotnet/Connection.h
```diff
@@ -33,8 +33,15 @@ inline HINSTANCE get_munique_client_library_handle()
 #else
 inline void* get_munique_client_library_handle()
 {
+    // Native AOT emits a platform-native shared object on Linux (.so), not the
+    // Windows .dll. Try the binary's own directory first (it is copied next to
+    // the executable by the build), then fall back to the loader search path.
     // Not const-qualified return: dlsym() takes a non-const void* handle.
-    static void* const handle = dlopen("MUnique.Client.Library.dll", RTLD_LAZY);
+    static void* const handle = []() -> void* {
+        if (void* h = dlopen("./MUnique.Client.Library.so", RTLD_LAZY))
```
Mosch0512: The comment says "Try the binary's own directory first", but `"./MUnique.Client.Library.so"` resolves against the **current working directory**, not the executable's directory. If the client is ever launched from a different CWD (a launcher, a desktop entry, a debugger), the `./` probe misses and you fall through to the bare name, which only works if the `.so` is on the loader search path.

If the intent really is "next to the executable", consider building the `.so` with an rpath of `$ORIGIN` (CMake `INSTALL_RPATH`/`BUILD_RPATH`) and just `dlopen("MUnique.Client.Library.so")`, or resolve the real exe dir (you already have `GetModuleFileNameW` -> `/proc/self/exe`) and `dlopen` an absolute path. As written the `./` line is CWD-dependent.

src/source/Core/Platform/WinApiShims.h
```diff
@@ -50,10 +119,21 @@ inline DWORD GetModuleFileNameW(HMODULE /*hModule*/, LPWSTR lpFilename, DWORD nS
...
 
 // Terminate the process (Win32 ExitProcess). Never returns, like the real one.
 [[noreturn]] inline void ExitProcess(UINT uExitCode) { exit(static_cast<int>(uExitCode)); }
 
+// Process / thread identifiers (Win32). Used only to tag temp file names, so
+// the POSIX pid / thread id serve the same uniqueness purpose.
+inline DWORD GetCurrentProcessId() { return static_cast<DWORD>(::getpid()); }
+inline DWORD GetCurrentThreadId()
+{
+    return static_cast<DWORD>(reinterpret_cast<std::uintptr_t>(pthread_self()));
```
Mosch0512: `pthread_self()` is an 8-byte value on 64-bit Linux (glibc `pthread_t` is an `unsigned long`); casting it down to a 32-bit `DWORD` discards the high half, so two threads can collide on the low 32 bits. It's only used to tag temp file names so the blast radius is small, but a collision would clobber another thread's temp file. Hashing the full value (e.g. fold the high and low halves together) before truncating would keep more entropy. Minor - flagging, not blocking.

.github/workflows/linux-build.yml
```diff
@@ -0,0 +1,92 @@
+name: Linux Build (x64)
+
+# Builds the full native Linux client (issue #462): the engine, the SDL/OpenGL
+# runtime, and the linux-x64 Native AOT network library (MUnique.Client.Library.so),
+# for both editor configurations. Replaces the old SDL-shell smoke test now that
+# the whole client compiles, links, and runs on Linux.
+
+on:
+  push:
```
Mosch0512, line 9: `on: push` + `on: pull_request` with no branch filter means that for any PR opened from a branch in this repo, both events fire and the whole matrix runs twice (the old `mingw-build-pr.yml` / `mingw-build.yml` split avoided this by scoping `push` to `main`). `windows-build.yml` has the same pattern. Consider `push: { branches: [main] }` (or a `concurrency:` group to cancel superseded runs) to avoid doubling Actions minutes on every PR push.

PR #502, opened 2026-06-16
[Fix] Click-through on the window close “X” (yellow one) and de-duplicate the handler

nolt:
# Fix click-through on the window close “X” (yellow one) and de-duplicate the handler

## Summary

The top-right “X” close glyph that sits on the shared item-frame
(`newui_item_back04.tga`) — used by the inventory, character, pet, NPC shops,
quests, storage, trade-style and several event windows — did not swallow the
click. Clicking the “X” fell through to the world, so the character **walked
toward the spot you clicked** instead of just closing the window. The clickable
area was also smaller than the visible glyph, so the close often “missed”.

On top of that, the exact same corner hit-test was **copy-pasted across ~36
window classes** (mostly as a `ptExitBtn1` block, sometimes `pClose` or an inline
`CheckMouseIn`), each with the same latent bug.

This PR introduces one shared handler, routes every window through it, and fixes
the click-through in a single place.

## Root cause of the click-through

World movement is gated in `ZzzInterface.cpp` by
`if (!MouseOnWindow && !g_pNewUISystem->CheckMouseUse())` and acts on the **raw**
`MouseLButtonPush` flag. The duplicated corner blocks called `Hide(...)` and
returned a consumed result, but never cleared `MouseLButtonPush`, so the press
still reached the movement code and the character walked. Returning “handled”
from the UI was not enough.

## What it does

- **New shared helper** `CNewUISystem::HandleFrameCornerClose(const POINT& winPos, DWORD dwKey)`:
  - hit-tests the corner glyph (box enlarged a little past the old `13×12` so the
    whole visible “X” is clickable — one place to tune for every window),
  - hides `dwKey`,
  - clears `MouseLButton` / `MouseLButtonPop` / `MouseLButtonPush` so the click
    can’t fall through to the world,
  - returns `true` when it handled the click.
- **Replaced the duplicated corner-close block in 36 windows** with a call to the
  helper, preserving each window’s own behavior (its `return true`/`return false`
  or no-return, side effects like `PlayBuffer`, the `m_bSellingItem` guard in the
  NPC shop, `else if` chains, and `m_Pos` / `m_ptPos` / `GetPos()`).
- **Added the missing handler** to two windows that rendered the frame (so they
  show the glyph) but had no corner close at all: the inventory extension (`K`)
  and the lucky-coin registration window.
- Removed a pre-existing dead variable (`ptExitBtn2`) left in the castle-siege
  window.

Net change: **−77 lines** across 38 files.

## Notes / intentional changes

- **Purchase-shop window** was normalized from `IsRelease` to `IsPress` so it
  matches every other window (and the helper). The original used release, which
  was an inconsistency rather than an intended difference.
- **Two windows are intentionally left as-is** because they already behave
  correctly and do not fit the `Hide(iface)` shape:
  - the trade-style window closes through its own `ProcessCloseBtn()`,
  - MU Helper reacts on hover and uses a different base frame.

## Testing

Built and run on the native Linux client. Verified live that the corner “X” now
closes the window **and the character no longer walks** for the inventory,
character, pet and chaos-machine windows (representative of the bare-call,
`return true` and `return false` groups). The remaining windows go through the
same helper, so they behave identically.

### DIFF
--- a/src/source/Guild/NewUIGuildInfoWindow.cpp
+++ b/src/source/Guild/NewUIGuildInfoWindow.cpp
@@ -189,17 +189,15 @@ bool SEASON3B::CNewUIGuildInfoWindow::UpdateMouseEvent()
 
 bool SEASON3B::CNewUIGuildInfoWindow::Check_Btn()
 {
-    POINT ptExitBtn1 = { m_Pos.x + 169, m_Pos.y + 7 };
-
     if (m_BtnExit.UpdateMouseEvent() == true)
     {
         g_pNewUISystem->Hide(SEASON3B::INTERFACE_GUILDINFO);
         m_EventState = EVENT_NONE;
         return false;
     }
-    if (SEASON3B::IsPress(VK_LBUTTON) && CheckMouseIn(ptExitBtn1.x, ptExitBtn1.y, 13, 12))
+    // Top-right corner close "X" (shared frame): hides + swallows the click.
+    if (g_pNewUISystem->HandleFrameCornerClose(m_Pos, SEASON3B::INTERFACE_GUILDINFO))
     {
-        g_pNewUISystem->Hide(SEASON3B::INTERFACE_GUILDINFO);
         m_EventState = EVENT_NONE;
         return false;
     }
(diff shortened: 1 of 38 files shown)


Mosch0512 (commented):
in PR Description you write "hit-tests the corner glyph (box enlarged a little past the old 13×12 so the whole visible “X” is clickable — one place to tune for every window),"

I checked the code its not actually the case not longer required or accidently missed?

nolt:
Yea, its no clonger the case. This feature is ready.

PR #529, opened 2026-07-16
fix: duplicate buff icon

eduardosmaniotto:
## Summary
**WSclient.cpp — ReceiveMagic / ReceiveMagicFinish**
- Register `eBuff_SwellOfMagicPower` (82) on cast for `AT_SKILL_EXPANSION_OF_WIZARDRY` / `_STR` / `_MASTERY`
- Unregister on expiry in `ReceiveMagicFinish`
- Matches the pattern every other buff skill in the file already follows

**WSclient.cpp — RegisterBuff / UnRegisterBuff**
- Remove the early-`return` when `g_IsBuffClass` returns `eBuffClass_Count`
- Guard `InsertBuffPhysicalEffect` / `InsertBuffLogicalEffect` / `ClearBuffPhysicalEffect` / `ClearBuffLogicalEffect` behind the class check instead
- Unknown-class buffs still enter `m_BuffMap` but skip effect handlers

**NewUIBuffWindow.cpp — NormalizeBuffState + BuffSort dedup**
- `NormalizeBuffState` maps 12 variant IDs to their 9 canonical bases:

| Variant IDs | Base | Chain |
|---|---|---|
| 135, 136 | `eBuff_Life` (8) | Swell Life |
| 138, 139 | `eBuff_SwellOfMagicPower` (82) | Expansion of Wizardry |
| 148, 149 | `eBuff_AddCriticalDamage` (5) | Critical Damage Increase |
| 143 | `eBuff_InfinityArrow` (6) | Infinity Arrow |
| 144 | `eDeBuff_Blind` (73) | Blind |
| 160 | `EFFECT_POISON_ARROW` (159) | Poison Arrow |
| 161 | `EFFECT_BLESS` (142) | Bless |
| 165 | `EFFECT_IRON_DEFENSE` (134) | Iron Defense |
| 167 | `EFFECT_BLOOD_HOWLING` (166) | Blood Howling |

- `BuffSort` uses two passes: pass one finds the highest variant present per chain, pass two skips lower-tier members
- Only the highest-tier icon renders; base buffs stay in `m_BuffMap` for stat calculations

**Before this PR:**
- MuHelper/manual casts of Expansion of Wizardry didn't recognize the buff as active
- Upgraded skills (STR / MASTERY / PROFICIENCY) showed duplicate icons in the buff window

**After this PR:**
- Base skill → correct base icon
- Upgrade skill → correct upgrade icon (never both)
- MuHelper detects the buff regardless of upgrade tier

## Related issues

NA

## Checklist

- [X] I have built the project locally (see [`docs/build/README.md`](docs/build/README.md)).

### DIFF
--- a/src/source/Network/Server/WSclient.cpp
+++ b/src/source/Network/Server/WSclient.cpp
@@ -3858,7 +3858,6 @@ void ReceiveMagicFinish(const BYTE* ReceiveBuffer)
     case AT_SKILL_ICE_STORM:
         UnRegisterBuff(eDeBuff_Freeze, o);
         break;
-        //  몬스터.
     case AT_SKILL_MONSTER_MAGIC_DEF:
         SetActionDestroy_Def(o);
         UnRegisterBuff(eBuff_Defense, o);
@@ -3881,6 +3880,11 @@ void ReceiveMagicFinish(const BYTE* ReceiveBuffer)
         break;
     case AT_SKILL_ALICE_THORNS:
         UnRegisterBuff(eBuff_Thorns, o);
+        break;
+    case AT_SKILL_EXPANSION_OF_WIZARDRY:
+    case AT_SKILL_EXPANSION_OF_WIZARDRY_STR:
+    case AT_SKILL_EXPANSION_OF_WIZARDRY_MASTERY:
+        UnRegisterBuff(eBuff_SwellOfMagicPower, o);
         break;
     case AT_SKILL_ALICE_BERSERKER:
     case AT_SKILL_ALICE_BERSERKER_STR:
@@ -4864,6 +4868,8 @@ BOOL ReceiveMagic(const BYTE* ReceiveBuffer, int Size, BOOL bEncrypted)
     {
         SetAttackSpeed();
         SetAction(so, PLAYER_SKILL_SWELL_OF_MP);
+
+        g_CharacterRegisterBuff(to, eBuff_SwellOfMagicPower);
 
         vec3_t vLight;
         Vector(0.3f, 0.2f, 0.9f, vLight);
@@ -15520,10 +15526,6 @@ void ClearBuffPhysicalEffect(eBuffState buff, OBJECT* o)
 
 void RegisterBuff(eBuffState buff, OBJECT* o, const int bufftime)
 {
-    eBuffClass buffclasstype = g_IsBuffClass(buff);
-
-    if (buffclasstype == eBuffClass_Count) return;
-
     if (!o)
     {
         return;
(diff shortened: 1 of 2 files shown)


Mosch0512 (commented):
Took a proper read through this. The core fix works. I checked the buff data file and `eBuff_SwellOfMagicPower` (82) is classed as a normal Buff, so the icon renders fine and MuHelper picks it up now. The dedup does what it says too, one icon per chain. Nice.

A few things I spotted while reading `BuffSort`. None of them break anything today, mostly stuff worth tightening while you're in here.

**1. The assert in BuffSort is now reachable**

This PR deliberately lets classless buffs into `m_BuffMap` (the "still enter m_BuffMap but skip effect handlers" part). That changes the meaning of this branch:

```cpp
else assert(!"SetDisableRenderBuff");
```

Before, the early return in `RegisterBuff` made sure nothing classless (`eBuffClass_Count`) could ever be in the map, so that `else` was basically dead code. Now that the early return is gone, if any classless buff ever lands on the Hero it'll hit the assert: abort in Debug, or silently draw nothing in Release while the entry just sits there. Swell itself is safe since it has a real class, but I'd swap that assert for a plain `continue`, because "classless buff in the map" is now a legit state rather than an impossible one.

This is the only one I'd actually push on, since the PR is what makes it reachable.

**2. Picking the top tier by enum id**

```cpp
if ((int)buf > (int)top[base]) top[base] = buf;
```

This picks the icon by "biggest enum value wins", which assumes higher id always means higher skill tier. True for most chains, but the crit damage one is backwards:

```
EFFECT_GREATER_CRITICAL_DAMAGE_MASTERED = 148
EFFECT_GREATER_CRITICAL_DAMAGE_EXTENDED = 149
```

So if a character ever has both up at once it'd show EXTENDED even though MASTERED is the top node. Edge case since normally only one variant of a skill is active, just flagging it.

**3. The `present[base]` check does nothing**

```cpp
if (present[base] && buf != top[base]) continue;
```

`present[base]` is always true by the time you reach this. Pass one already set it for every buff that isn't `eBuffNone`, and pass two walks the same list with the same `eBuffNone` skip, so the condition is really just `buf != top[base]`. Not wrong, just reads like there's a false case that can't actually happen. Could drop the `present[base] &&`.

**4. The two passes treat disabled buffs differently**

Pass one doesn't skip `SetDisableRenderBuff` buffs, pass two does. Harmless right now because none of the disabled ids are in any of the normalize chains. But if one ever gets added to a chain, a disabled buff could win `top[base]` in pass one and then get skipped in pass two, which would hide the real icon for that whole chain. Adding the same `SetDisableRenderBuff` skip to pass one would keep them in sync.

Like I said, nothing here is a blocker. Mainly the assert one. The rest is your call.

PR #530, opened 2026-07-20
fix: apply in-game resolution changes through SDL

alexyabar-grimerian:
## Problem

The options-window **Resolution** combo did not apply in windowed mode. The window is SDL-owned, but `ApplyResolution` drove the Win32 `SetWindowPos` / `ChangeDisplaySettings` path on `g_hWnd`. SDL pins the min/max tracking size of a non-resizable window, so the raw `SetWindowPos` was clamped straight back to the current size — changing the resolution did nothing. It only "took" when a windowed/fullscreen toggle happened to reset the window style first, which is why it seemed to work intermittently.

## Fix

Route `ApplyResolution` through `MuApplyWindowResolution` (`SDL_SetWindowSize` / `SDL_SetWindowFullscreen`) on **every** platform, and compile `MuApplyWindowResolution` on Windows too (it was `#ifndef _WIN32`). It resizes regardless of the resizable flag and drives `HandleWindowResize` synchronously, so the config is saved with the size the user actually got. This also drops ~70 lines of Win32-specific window juggling from `ApplyResolution`.

Follow-ups from review:

- The windowed/fullscreen checkbox toggle also routes through SDL now instead of raw Win32 (`ChangeDisplaySettings` / `SetWindowLongPtr`), so SDL's window state never diverges from the OS state.
- `MuApplyWindowResolution` settles the request with `SDL_SyncWindow` and reads back the size the window actually got, so a coerced fullscreen request (closest-mode pick or borderless fallback) is what gets persisted and shown in the combo — `config.ini` records what happened, not what was asked for. The toggle path persists mode + resulting size together, after the apply.

Note: in windowed mode the window is now **re-centered** on the display after a resolution change (`SDL_WINDOWPOS_CENTERED`), replacing the old fixed position.

## Testing

Windows x64, windowed mode: changing the Resolution combo 1024×768 → 1280×720 → 800×600 resizes the window live each time (verified by measuring the client rect), re-centers the window, and persists to `config.ini`. Relaunch opens at the saved size.

Fullscreen pass (Windows x64): toggling windowed ↔ fullscreen in both directions switches the display mode and keeps the combo and `config.ini` in agreement; changing resolution while fullscreen works for modes the display supports (1280×720, 1280×1024). Coercion case: requesting 1600×1200 on a panel without that mode — SDL fell back to 1920×1080, the combo snapped to 1920×1080, and `config.ini` recorded 1920×1080. Relaunch honors the saved state.

Refs #462.

### DIFF
--- a/docs/options-window.md
+++ b/docs/options-window.md
@@ -23,6 +23,12 @@ What was added on top of that:
   configured maximum.
 - **Resolution combo-box clicks no longer leak through** to the checkboxes
   behind the dropdown.
+- **Resolution changes go through SDL** (PR
+  [#530](https://github.com/sven-n/MuMain/pull/530)), which makes the live
+  apply actually work in windowed mode on Windows. Windowed resizes re-center
+  the window on the display. Fullscreen picks the closest real display mode
+  (or borderless desktop when nothing matches), and the combo and
+  `config.ini` record the mode that actually resulted, not the one requested.
 
 ## What it stores in `config.ini`
 
--- a/src/source/App/Platform/Windows/Winmain.cpp
+++ b/src/source/App/Platform/Windows/Winmain.cpp
@@ -1033,12 +1033,15 @@ namespace
     }
 }
 
-#ifndef _WIN32
-// Portable resolution change (issue #462). The Win32 path in ApplyResolution()
-// is entirely g_hWnd-gated and drives a synchronous WM_SIZE, neither of which
-// exists here, so the option window calls this instead. SDL owns the window;
-// resize it and apply the new dimensions immediately (SDL also posts a resize
-// event, but callers Save() config right after and must see the new size).
+// Resolution change through SDL (issue #462). SDL owns the window on every
+// platform, so resize it via SDL rather than the OS. The old Windows path in
+// ApplyResolution() drove Win32 SetWindowPos/ChangeDisplaySettings on g_hWnd,
+// which fought SDL: it pins the min/max tracking size of a non-resizable
+// window, so a raw SetWindowPos was clamped back and the resolution never
+// changed unless a windowed/fullscreen toggle reset the style first.
+// SDL_SetWindowSize resizes regardless of the resizable flag and drives the
+// same HandleWindowResize update synchronously, so callers can Save() config
+// right after and see the new size.
 void MuApplyWindowResolution(unsigned int width, unsigned int height, bool windowed)
 {
     if (!g_sdlWindow || width == 0 || height == 0) return;
@@ -1067,7 +1070,6 @@ void MuApplyWindowResolution(unsigned int width, unsigned int height, bool windo
 
     HandleWindowResize(w, h);
 }
-#endif // !_WIN32
 
 MSG MainLoop()
 {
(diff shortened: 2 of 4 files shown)


Mosch0512 (changes requested):
Nice cleanup, routing everything through SDL is clearly the right call and getting rid of all that Win32 juggling makes this way easier to reason about. The windowed fix looks good to me.

One thing I want fixed before this goes in though. In the fullscreen path MuApplyWindowResolution calls HandleWindowResize with the width/height we *asked* for, and then ApplyResolution saves WindowWidth/WindowHeight right after. But SDL can pick a different mode with SDL_GetClosestFullscreenDisplayMode, or fall back to borderless desktop when nothing matches. When that happens config.ini and the combo end up showing a resolution the monitor never actually switched to.

The old code guarded against this by snapping the combo back on a rejected mode so the UI wouldn't lie, and that guard is gone now. Can we read the size back from the window after the switch instead of trusting the requested values? Something like SDL_GetWindowSizeInPixels after SDL_SetWindowFullscreen, and feed that into HandleWindowResize / the Save. That way what we persist matches what actually happened.

Also the testing notes only cover windowed mode, but the fullscreen path changed the most here. Could you give it a pass too? Toggle windowed to fullscreen and back, change resolution while fullscreen, and check that both the actual display and config.ini agree.

Small stuff, not blocking:
- extern HWND g_hWnd in NewUIOptionWindow.cpp is dead now, only shows up in a comment, might as well drop it with the rest
- worth mentioning in the PR that windowed mode now centers the window instead of the old fixed position, just so it's not a surprise

alexyabar-grimerian:
Thanks for the careful pass — the fullscreen point was exactly right, and fixing it properly surfaced one more spot with the same problem. All four items addressed in 2a80c44 + 3d9c4df:

**Read-back of the actual size** — `MuApplyWindowResolution` now settles the request with `SDL_SyncWindow` (mode switches are asynchronous on some window managers) and then feeds `SDL_GetWindowSize` into `HandleWindowResize`, so `WindowWidth/Height` — and everything persisted from them — reflect what the window really got, whether the closest-mode pick differed from the request or the borderless fallback kicked in.

One deliberate detail: `SDL_GetWindowSize` (logical size) rather than `SDL_GetWindowSizeInPixels`. The event path already drives `HandleWindowResize` with logical coordinates (`SDL_EVENT_WINDOW_RESIZED`'s `data1`/`data2`) and the window doesn't request `SDL_WINDOW_HIGH_PIXEL_DENSITY`, so logical is the unit the rest of the code speaks — on a hiDPI display the pixel variant would disagree with what the resize event later reports.

**Combo snap** — restored as `SyncResolutionComboToWindow()`, called after every apply: if the resulting size matches a listed mode, the combo snaps to it, so the UI never claims a resolution the display didn't adopt. If the result isn't in the list (borderless desktop on an unlisted panel size), the selection is left alone but `config.ini` still records the real size. `FindCurrentResolutionIndex` and the snap share one table scan now.

**Toggle path** — while adding the read-back I noticed the windowed/fullscreen toggle had the same bug from the other side: it saved config *before* applying the mode, so a coerced fullscreen size never reached disk and the "display and config.ini agree" check would have failed there too. It now applies first, then persists mode + actual size together (and the block moved out of `UpdateMouseEvent` into `ApplyWindowModeToggle()`, following the file's handler pattern).

**Dead `extern HWND g_hWnd`** — dropped, and the PR description now mentions the windowed-mode centering change.

**Fullscreen testing** (Windows x64): toggling windowed ↔ fullscreen both ways switches the display and keeps the combo and `config.ini` in agreement; resolution changes while fullscreen work for modes the display supports (1280×720, 1280×1024). Coercion case: requesting 1600×1200 on a panel without that mode — SDL fell back to 1920×1080, the combo snapped to 1920×1080, and `config.ini` recorded 1920×1080. Relaunch honors the saved state.

Mosch0512 (approved):

PR #540, opened 2026-07-26
Offer the chat commands of the player in a window

### LINKED ISSUE #539: Show available (chat) commands in a UI instead of requiring /list

sven-n:
Closes #539, the client side of MUnique/OpenMU#847.

Until now the only way to find a chat command was `/list`, which the server
answers with one blue system message per command. The info window keeps the last
few and has no scrollback, so with ~60 commands the list is gone before it can be
read - and even then, the player has to know that `/list` exists and type every
command by hand.

The server can now describe its commands (MUnique/OpenMU#852), so the client asks
for them and offers them.

## 1. Receiving the commands

- `ChatCommandCatalog` under `src/source/GameLogic/Commands/` parses the
  `AvailableChatCommand` message: command, localized name and description, the
  required character status, and one entry per parameter with its name, short
  name, kind and accepted values.
- `ProcessPacket` gets a short `0xF5` case which only delegates - the parsing
  stays out of `WSclient.cpp`, in line with the extraction work going on there.
- The request goes out once the character entered the world, guarded so it
  doesn't repeat on every map change.
- Executing needs no own message: `SendPublicChatMessage` is enough, because the
  server routes a line starting with a slash to the command instead of
  broadcasting it. The line is also put into the chat history, so the arrow keys
  repeat it without opening the window again.

The packet package moved to **0.9.10**, which is the version carrying these
messages. Its version was spelled out in **17** places in
`MUnique.Client.Library.csproj` - the `PackageReference` and every XSLT input
path - so it's a single `OpenMuPacketsVersion` property now. They all have to
move together, otherwise the transformation reads a different XML than the
restore produced. `SendChatCommandListRequest()` is generated from it, not hand
written.

## 2. The window

`CNewUIChatCommandWindow` has the width (190) and the frame composition of the
command window, and shows one page at a time. It opens with **J**, or with the
new **Special** button in the **D** command window - both land on the first page.

**Page 1 - the commands.** By their name only, because that is all that fits at
this width. Three dots mark the ones which take values, a leading `*` marks a
favourite, and favourites are listed first.

- **A command with no parameters is sent with one click** and the window closes.
  That's 13 of the currently 65 commands - the event starters, `/hide`,
  `/unhide`, `/reset` and their kind - including the "start Blood Castle on
  demand" case which started this whole topic in Discord.
- A command with parameters opens page 2, including the ones whose parameters are
  all optional: they still need a way to be filled in.

**Page 2 - the values.** Description, then one box per parameter.

- Parameters which only accept a set of values (booleans included, the server
  sends `0|1` for those) are **cycled by clicking**; one step past the last value
  clears it again, which is how an optional parameter is left out.
- Everything else is **typed**, and a parameter of a numeric kind only accepts
  digits.
- A required parameter without a value is marked and keeps the command from being
  sent, since it would only earn an error message from the server.
- Two actions below the values: **add to / remove from favourites**, and **save
  as template**.

**Page 3 - the templates.** A template is a command with the values which were
entered for it, shown as the command line it produces. Clicking it sends that
line, the `x` at its right removes it.

Favourites and templates are stored per installation in the `[ChatCommands]`
section of `config.ini`, not per character - the commands a player reaches for
don't change when the character does.

`BuildCommandLine` writes the `shortName=value` form when the parameters have
short names. That's deliberate: the server switches to named parsing as soon as
the line contains an equals sign, which is what makes the optional parameters
skippable.

## Fixed along the way

- **Number fields accepted letters.** The field now gets `UIOPTION_NUMBERONLY`
  for numeric parameters.
- **An empty value left a stray glyph behind.** `CUIRenderTextOriginal::RenderText`
  measures a `"0"` placeholder for an empty string and still draws when a box
  width is given, so the window must not ask it to draw one.
- **The background didn't cover the dialog.** It was 380 wide with a frame which
  is drawn for 190; at the width the frame is made for, it composes the same way
  it does for the inventory and the command window.
- **The window was never told when it opens or closes.** `INTERFACE_COMMAND_LIST`
  had no branch in the show/hide dispatch, so `OpenningProcess` and
  `ClosingProcess` never ran. It's in group A now, like the command window.

## What this doesn't do yet

- No search box - the list scrolls with the wheel and the arrow keys.
- Only the English resource strings are filled; the other languages fall back to
  them.

Semantic parameter kinds - "this is an item, this is a monster" - are
MUnique/OpenMU#849. Once the server sends those, the item and monster rows can
offer a picker with the names the client already knows.

## Verification

Built on Linux (`cmake --build`, exit 0) after every step, including after the
merge with `main`, so the C++ compiles. What I could **not** do is run the client
against a server, so the exact pixel positions of the rows and the click regions
are still unverified - that part needs your eyes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MSpK6jkyF8ZS5nYXyGwYxA)_


sven-n:
Screenshot:
grafik

After doing a quick test, I found the following issues:
* The list request is only fired after a map change. It does not load right after login/world entering and therefore the dialog cannot open.
* There is no send button visible.
* I think the list is not scrollable and doesn't show all commands. Suggestions:
  * Make it bigger so that more height is used.
  * Add a vertical scroll bar.
  * Use more height for the initial list. Make space for the description, inputs and send button when one command is selected.
* The background is black at the right side of the dialog.
* The description of the command is cut off when it's too long. Add a line break instead.
* The dialog has no nice borders like other mu client dialogs.
* The inputs are not rendered
* Add the 'J' key binding to the F1 help text.

Additionally, I suggest the following improvements:
* Allow to select commands as favourites (star icon or similar).
* Allow to save certain commands as templates with a name, list them above the command list. It should be possible to remove them, too. When a template is selected, the values are pre-filled and the user just has to press the send button.

sven-n:
Thanks for testing it - the screenshot makes the state of it obvious.

## Fixed

**The list was never requested after login.** The request went out in
`ReceiveTeleport`, which only runs on a map change, so after entering the world
there was nothing to show and the window refused to open. It's in the handler of
the join map server message now, which is the one that runs when a character
enters. The catalog is also reset there, so the commands of a previously played
character don't survive a character switch.

**`J` is listed in the F1 help** now, next to the other keys.

## Diagnosed, not fixed yet

The missing send button, the black area on the right, the missing borders and
most likely the invisible inputs are all **one cause**: the window borrows image
constants from `CNewUIMessageBoxMng` and never loads any bitmaps of its own.
Other dialogs have a `LoadImages`/`UnloadImages` pair (e.g. `CNewUIGensRanking`)
and build their frame from the corner and edge pieces. Mine just stretches one
constant across the whole window, which is why the right side is black and why
the button doesn't appear - the texture behind that constant isn't loaded when
my window draws.

So the fix isn't a set of small patches, it's giving the window a proper frame
the way the other dialogs do it, and that naturally covers:

- the borders,
- the background,
- the send button,
- and the inputs, which I need to check separately: `CUITextInputBox::Render` may
  not be enough on its own.

## Still open from your list

- Bigger window, more height for the list, and a vertical scroll bar
  (`CNewUIScrollBar` exists and `CNewUIGensRanking` is the only current user, so
  there's a pattern to follow).
- Splitting the layout: list on top, description, inputs and send button below,
  appearing when a command is selected.
- Wrapping long descriptions instead of cutting them off.

## Your two suggestions

Favourites and named templates both make sense to me, especially templates -
`/item` with a set of values is exactly the thing you'd want to keep. Both need
somewhere to persist per character, so I'd rather do them as a follow-up once the
window itself looks right, unless you want them in here.

## One thing to keep in mind

I can build the client here but not run it against a server, so everything about
layout, colors and hit areas is unverified on my side - which is exactly how the
issues above got through. Screenshots like yours are the only feedback loop I
have for the visual part, so please keep them coming.

---

sven-n:
I think it's (almost) finished now :)
Some commands probably need more metadata, e.g. '/post'.

grafik

PR #541, opened 2026-07-27
Restrict advanced class creation

konstantin19881220-maker:
## What changed

- Disabled Dark Lord, Summoner, and Rage Fighter on the character creation screen.
- Added a request-time guard so restricted classes cannot be submitted even if the UI state is bypassed.
- Documented the classes available for new character creation.

## Why

This client should allow creation of Dark Knight, Dark Wizard, Fairy Elf, and Magic Gladiator only.

### DIFF
--- a/docs/character-creation.md
+++ b/docs/character-creation.md
@@ -1,0 +1,10 @@
+# Character creation
+
+The character creation screen allows new characters from the following classes:
+
+- Dark Knight
+- Dark Wizard
+- Fairy Elf
+- Magic Gladiator
+
+Dark Lord, Summoner, and Rage Fighter remain visible but disabled and cannot be submitted for creation from the client.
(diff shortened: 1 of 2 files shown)


Mosch0512:
i would like this with some sort of buildgameconfig file or sth that has what characters should be built into the game or not.
Because as i get what you are trying to do is that this main cant create these classes anymore right?

jkok25:
Currently, OpenMU is configured to unlock character creation (however, the client can bypass this - this is a bug that needs to be fixed in the OpenMU project - the client should only allow server-side creation of unlocked characters):
https://github.com/MUnique/OpenMU/tree/master/src/GameLogic/PlugIns/UnlockCharacterClass
UnlockSummonerAtLevel1
UnlockRageFighterAtLevel150
UnlockMagicGladiatorAtLevel220
UnlockDarkLordAtLevel250

PR #542, opened 2026-07-27
Hide restricted character classes

konstantin19881220-maker:
## What changed

- Removed Dark Lord, Summoner, and Rage Fighter from the visible character creation choices.
- Kept the request-time guard so hidden classes cannot be submitted if the UI is bypassed.
- Documented the classes available for new character creation.

## Why

This alternative version should show only Dark Knight, Dark Wizard, Fairy Elf, and Magic Gladiator as creation options.

### DIFF
--- a/docs/character-creation.md
+++ b/docs/character-creation.md
@@ -1,0 +1,10 @@
+# Character creation
+
+The character creation screen allows new characters from the following classes:
+
+- Dark Knight
+- Dark Wizard
+- Fairy Elf
+- Magic Gladiator
+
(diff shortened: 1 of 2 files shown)


jkok25:
Currently, OpenMU is configured to unlock character creation (however, the client can bypass this - this is a bug that needs to be fixed in the OpenMU project - the client should only allow server-side creation of unlocked characters):
https://github.com/MUnique/OpenMU/tree/master/src/GameLogic/PlugIns/UnlockCharacterClass
UnlockSummonerAtLevel1
UnlockRageFighterAtLevel150
UnlockMagicGladiatorAtLevel220
UnlockDarkLordAtLevel250

PR #551, opened 2026-08-08
feat (render) : Disable KHR_debug Callback for Performance

janblade:
# Disable KHR_debug Callback for Performance

## Description
This PR gates the OpenGL `KHR_debug` callback setup behind a preprocessor macro (`ENABLE_GL_KHR_DEBUG_CALLBACK`), effectively disabling it by default in all build configurations, including `_DEBUG`.

## Motivation
The `KHR_debug` callback was initially introduced as a temporary diagnostic tool during the DXP-08 OpenGL Core Profile transition to help attribute errors synchronously. However, using `GL_DEBUG_OUTPUT_SYNCHRONOUS` alongside heavy `dbghelp.lib` stack walkbacks (`CaptureStackBackTrace`) caused massive stalls and severely degraded rendering performance during debug development iteration. 

Since the core profile port is now functionally verified and we no longer need constant synchronous error attribution, this diagnostic overhead has been disabled. The mechanism is preserved as a "nuclear option" (`#define ENABLE_GL_KHR_DEBUG_CALLBACK 1`) for future debugging if severe GL rendering issues arise.

## Changes Made
- Introduced the `ENABLE_GL_KHR_DEBUG_CALLBACK` macro in `Winmain.cpp`.
- Gated the `SDL_GL_CONTEXT_DEBUG_FLAG` attribute during SDL context creation to avoid unnecessary driver debug overhead.
- Guarded `GLDebugCallback` definitions, `LogSymbolizedStack`, and `SDL_GL_GetProcAddress` registration logic behind the macro.
- The `ENABLE_GL_KHR_DEBUG_CALLBACK` toggle is positioned near the diagnostic functions it controls for graceful toggling when needed.

### DIFF
--- a/src/source/App/Platform/Windows/Winmain.cpp
+++ b/src/source/App/Platform/Windows/Winmain.cpp
@@ -1483,7 +1483,13 @@ void UpdateResolutionDependentSystems()
     CUIMng::Instance().RepositionSceneUI();
 }
 
-#if 0 // Disabled KHR_debug callback due to severe Debug mode performance degradation (GL_DEBUG_OUTPUT_SYNCHRONOUS stalling)
+#ifdef _DEBUG
+// Set to 1 to enable GL_KHR_debug callback logging (DXP-08 diagnostic soak mode).
+// Disabled (0) by default because GL_DEBUG_OUTPUT_SYNCHRONOUS and stack symbolization cause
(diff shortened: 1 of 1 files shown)


nitoygo:
Confirmed fps drop fixed

Issue #552, opened 2026-08-09
[Graphics] Performance & FPS Audit (AI-Generated)


nitoygo:
# Performance & FPS Audit — Rendering, Computation, and Frame Pacing

**Scope:** current state of the MuMain client codebase (SDL3 + OpenGL 3.3 Core Profile), covering rendering, CPU computation, GPU/texture memory, threading, and frame-pacing/FPS mechanics. Focus is on issues with a measurable real-world impact on frame rate, frame-time consistency, loading stalls, or responsiveness — especially on older or weaker hardware (integrated GPUs, low-core-count CPUs, limited VRAM, slower disks).

**Method:** targeted code investigation across the rendering pipeline, CPU hotspots, texture/GPU-memory management, and the main loop/frame-timing implementation, with severity and confidence explicitly assessed per finding rather than assumed from pattern-matching alone.

**Confidence levels used below:**
- **Confirmed** — traced directly through the code, mechanism and trigger condition both verified.
- **Likely risk** — mechanism confirmed, but real-world magnitude depends on data/workload not fully knowable from source alone.
- **Requires profiling** — the code pattern exists, but whether it's actually a bottleneck at realistic scale needs runtime measurement before acting.

Status: **findings only, not yet fixed.** Checkboxes below are meant to be pasted into a GitHub issue as trackable subtasks.

---

### High — confirmed, reproducible, and the clearest wins

- [x] **Frame-rate cap silently fails to engage on vsync rejection.** The engine's vsync-capability check is hardcoded to report success rather than actually probing the driver, and the frame-rate target is set to "uncapped" unconditionally in the same code path — regardless of whether `SDL_GL_SetSwapInterval(1)` actually succeeded. If a driver silently rejects vsync (common on laptops with hybrid/Optimus GPUs, some Linux/Mesa drivers, or borderless-fullscreen compositor bypass), the render loop has no fallback cap at all and pins one CPU core at 100%. *(`src/source/App/Platform/Windows/Winmain.cpp` vsync setup block; `src/source/Render/Textures/ZzzOpenglUtil.cpp` `InitVSync()`/`EnableVSync()`)* — Confidence: Confirmed.
[...]

PR #554, opened 2026-08-09
fix: fps regression measures - reduce redundant GPU upload traffic

nitoygo:
#### 2. `BoneUBO`'s dedup cache existed but was silently broken

Both UBOs now dedup by comparing **actual content**, not proxy signals (pointer/version):
- **`BoneUBO::UploadBones()`**: dropped the `version` parameter; now `memcmp`s the incoming bone-matrix bytes against a cached snapshot of the last upload before repacking/uploading. Unconditionally correct — no aliasing is possible when comparing real bytes — and cheap (comparing up to 9.6KB in cache is far cheaper than the GPU upload it skips).
- **`GlobalUBO::Upload()`**: `memcmp`s the packed 272-byte uniform buffer against the last upload before writing; skips the GPU call entirely when nothing changed. Safe unconditionally — `GlobalUBO`'s fields are the singleton's own persistent state, not a shared/reused external buffer, so there's no aliasing hazard analogous to `BoneUBO`'s.
- Removed the now-fully-dead `g_BoneTransformVersion` global and its unconditional increment in `SetActiveBoneTransform()`.
- Updated `docs/GPU Skinning/gpu-skinning-architecture.md` and `gotchas-and-patterns.md`, which had documented the old (broken) version-stamp scheme as the correct approach — left as-is, they'd have actively misled the next person who touched this code.

### 1. `GlobalUBO` had no dirty-checking at all
Every `SetModel`/`PushModel`/`PopModel`/`SetView`/`SetProj`/`SetTime` call triggered an unconditional GPU buffer upload — including pure *restores* like `PopModel()`, which re-uploaded a matrix that was already GPU-resident moments earlier. Traced to `ZzzCharacter.cpp`'s per-character `PushModel(); SetModel(...); ...; PopModel();` sequence: **2 avoidable GPU uploads per rendered character per frame**, with no equivalent cost under FFP (`glPopMatrix` was a free CPU-only operation).

### 2. `BoneUBO`'s dedup cache existed but was silently broken
`BoneUBO::UploadBones()` was designed to skip re-uploading a character's bone palette when consecutive meshes (body, armor, weapon, wings) share the same skeleton — but `SetActiveBoneTransform()` unconditionally bumped a global version counter on *every* call, regardless of whether the underlying pointer actually changed. Since the dedup check required an exact `(pointer, version)` match, it **never fired** for the common multi-mesh-per-character case: every equipped piece re-uploaded the full 12.8KB bone-matrix buffer redundantly instead of once per character.

### Important correction made mid-investigation
The obvious first fix for #2 — "only bump the version when the pointer changes" — was traced against the actual call graph and found to be **unsafe**, not just insufient. Weapons, wings, effects, and non-skeletal world objects all render through a single shared, fixed-address scratch buffer (`BoneTransform[MAX_BONES][3][4]` in `ZzzBMD.cpp`). Rendering a weapon, then wings, calls `Animation()`+`Transform()` on that *same address* with *completely different content* both times — exactly the scenario a nearby existing save/restore hack already exists to guard against (its comment cites a prior real bug: "Elf and Dark Lord rendering corruption in the character selection screen"). A pointer-only check would have silently reintroduced an equivalent stale-skeleton bug. This was caught before implementation by tracing every call site rather than trusting the first plausible fix.

## Related issues

https://github.com/sven-n/MuMain/issues/552
This PR addresses two *confirmed, redundant* sources of GPU upload traffic introduced by the migration. It does not address the other already-documented contributors to renderer overhead (no draw-call batching across terrain/models/UI/particles, `RHI::DrawIndexed` unimplemented) — those are separate, already-tracked findings from the broader performance audit, not fixed here.

## Checklist


janblade:
the BoneUBO fix is broader (and costlier per-call) than the actual bug warrants. Cheapest real fix would be to save/restore g_BoneTransformVersion alongside the pointer at the one broken call site, rather than switching the whole path to full-content comparison

I'd cherry-pick the 3 small fixes and treat the dedup commit as a decision point, not a given.

PR #557, opened 2026-08-11
fix(render): Incorrect tile resolution. Fixes #556

### LINKED ISSUE #556: Incorrect terrain texture resolution
After the latest updates (not sure which one), I noticed that the terrain got an incorrect texture, as you can see comparing the before and after:

Before:
Image

Current version:
Image

janblade:
## Summary

Terrain tiles rendered visibly larger/blurrier than the pre-FFP baseline (screenshotted on
Noria). `TerrainShader`'s GPU fast path computed ground-tile UV with a single hardcoded
constant (`a_Pos.xy * 0.0025`), but terrain tile bitmaps in this tree are not uniformly sized
(64/128/256px all present — confirmed by decoding the actual `.OZJ` assets). A shared constant
can only ever be correct for one texture resolution; Noria's tiles are 256px and rendered
stretched under the old value.

Fixes it by deriving the UV scale **per tile, per texture** in the vertex shader, mirroring the
legacy fixed-function `FaceTexture()`'s own formula (`Width = 64.f/b->Width`) instead of a baked-in
constant:

- `TerrainShader.h`/`.cpp`: new `SetUVScale(baseX, baseY, overlayX, overlayY)`, two new
  `u_BaseUVScale`/`u_OverlayUVScale` uniforms (dirty-checked like the existing `SyncAlphaRef`).
- `ZzzLodTerrain.cpp`: `RenderTerrainFace()`'s shader branch now computes the scale from each
  tile's actual bound texture (`64.f/b->Width`, `64.f/b->Height`, divided by `TERRAIN_SCALE`)
  right where `b1`/`b2` are already resolved for `SetBaseTexture`/`SetOverlayTexture`.

Fixes #556 

## Test plan

- [x] Debug build, clean incremental compile + link
- [x] In-game: Noria ground tiles visually compared against `perf-baseline-pre-ffp` — confirmed
      matching size (owner-verified)
- [x] Spot-check a map using a different tile resolution than Noria's 256px (fix is now
      data-driven per texture, so it should generalize, but not yet directly observed elsewhere)

### DIFF
--- a/src/source/Render/Shaders/TerrainShader.cpp
+++ b/src/source/Render/Shaders/TerrainShader.cpp
@@ -28,6 +28,7 @@ typedef GLuint (APIENTRY* PFNGLGETUNIFORMBLOCKINDEXPROC)(GLuint, const GLchar*);
 typedef void   (APIENTRY* PFNGLUNIFORMBLOCKBINDINGPROC)(GLuint, GLuint, GLuint);
 typedef void   (APIENTRY* PFNGLUNIFORM1IPROC)(GLint, GLint);
 typedef void   (APIENTRY* PFNGLUNIFORM1FPROC)(GLint, GLfloat);
+typedef void   (APIENTRY* PFNGLUNIFORM2FPROC)(GLint, GLfloat, GLfloat);
 typedef void   (APIENTRY* PFNGLACTIVETEXTUREPROC)(GLenum);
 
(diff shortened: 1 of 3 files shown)


janblade:
image

Left side post FFP retirement
Right side pre-FFP retirement baseline

Issue #559, opened 2026-08-12
[Animation] Chaotic Diseier - Skill Effect Draw Distance. FPS reduced by 50%


janblade:
Skill draws up to the edge of the map.
Zoomed out F10 to verify.

I think draw distance should be culled up to the actual skill range only

With high agi, skill casting Chaotic Diseier makes particles linger longer than expected.
with particles stacking up, performance is greatly reduced.
After 5seconds continuous skill casting, FPS is reduced by 50%

janblade:
<img width="1003" height="679" alt="Image" src="https://github.com/user-attachments/assets/01e5c243-ab39-480f-9aa8-ed9127131a06" />

PR #560, opened 2026-08-12
perf(render): GLP series — Core Profile FPS regression investigation and fixes

janblade:
## Summary

The `GLP-xx` ("glperf") series: an investigation and fix pass for a Core Profile FPS regression reported on weak/integrated hardware (**Intel HD 530 / i5-6500 / 8GB**) after the client's fixed-function-pipeline retirement — *"older version 100-60fps, newer version 40-23fps"* at the same in-game location.

25 commits across five phases: instrumentation, cheap per-call removals, killing the per-draw uniform allocation, and collapsing the draw count. Every performance claim below is a counter reading or a frame-time measurement, not an estimate.

---

## Measured result

Two `$glstats` captures in Lorencia. The delta between them is exactly five commits: **GLP-25** (IR ring growth), **GLP-19** (IR cross-`Begin`/`End` batching), **GLP-24's GPU query ring fix**, and **GLP-29 2a/2b/2c** (particle/sprite vertex-generation CPU).

### Frame level

| Metric | Before | After | Δ |
|---|---|---|---|
| Average FPS | 111.3 | 123.4 | **+10.9%** |
| **1% low** | **47.0** | **96.4** | **+105%** |
| Slowest frame | 30.9 fps | 94.3 fps | **+205%** |
| Frame time | 8.99 ms | 8.10 ms | **−9.9%** |
| Max FPS | 149.7 | 150.3 | flat |

The 1% low and slowest-frame numbers are the ones that matter here — the average moved ~11%, but frame-time consistency roughly doubled.

### Draw calls per pass

| Pass | Before | After | Δ |
|---|---|---|---|
| Terrain | 46 | 17 | −63% |
| Objects | 374 | 374 | — |
| Chars | 157 | 159 | +1% |
| Sprites | 150 | 31 | **−79%** |
| Particles | 150 | 146 | −3% |
| Joints | 25 | 5 | **−80%** |
| UI | 67 | 52 | −22% |
| Other | 90 | 92 | +2% |
| **Total** | **1,117** | **933** | **−16.5%** |

### Frame GL totals

| Counter | Before | After | Δ |
|---|---|---|---|
| GL calls | 3,977 | 3,672 | −7.7% |
| Draw calls | 1,117 | 933 | −16.5% |
| Buffer updates | 1,315 | 1,121 | −14.8% |
| — of which orphans | 5 | 1 | −80% |
| Program binds | 98 | 101 | +3% |
| Texture binds | 602 | 650 | +8% |
| Uniform writes | 537 | 551 | +2.6% |

Texture binds went **up** 8%. Batching changes which binds the dirty-check cache can suppress; it was not investigated because the frame-time result was positive regardless.

Elsewhere in the series, on their own like-for-like captures: terrain draw calls **1,226 → 48** (25.5×) from texture-pair bucketing (GLP-16), uniform-buffer orphans **912 → 4** from the ring allocator (GLP-09), and IR streaming-ring orphans **21 → 1** (GLP-25).

---

## Reading the numbers honestly

Four caveats a reviewer cannot infer from the screenshots, and I would rather state them than have them found:

1. **Not verified on the target hardware.** Every measurement in this series is from an **RTX 5070 Ti**. The Intel HD 530 machine that motivated the work has never run any of it. The series is internally consistent and the wins are real on the dev box, but the claim it was written to satisfy is unproven.
2. **These two captures are Debug builds**, and they are not perfectly like-for-like — Lorencia (131,129) vs (151,129), 5 vs 6 active characters. GLP-29's win in particular is overwhelmingly MSVC Debug instrumentation: on Release the same change moved the frame 5.99 → 5.92 ms, i.e. nothing. Treat the table as directional.
3. **The GPU-ms column is not comparable across these two builds.** `e15814af` fixed `AdvanceGpuTimers()` reading the ring slot the current frame had just written, which stranded late-submitted passes at 0.00 permanently. `Joints` GPU 0.00 → 1.66 and `Chars` 0.12 → 1.21 are the *measurement* being repaired, not a regression.
4. **Instantaneous FPS (118.1 → 111.2) is a single-frame sample** and should be ignored; average and 1% low are the meaningful figures.

---

## What is in the series

| Phase | Milestones |
|---|---|
| 0 — Ground truth | GLP-01 (GL call counters + `GL_TIMESTAMP` pass timers, `$glstats` overlay) |
| 1 — Cheap removals | GLP-03/04/05/06 (sampler uniform rewrites, terrain water-flag dirty-check, active-texture-unit cache, per-draw VAO calls) |
| 2 — Per-draw allocation | GLP-08 (GL context version chain + `RHI::Caps` probe), GLP-09 (ring-buffer UBO allocator), GLP-10 (content dirty-check), GLP-11 (bone palette as 3×vec4 affine rows) |
| 4 — Draw count | GLP-16 (terrain texture-pair bucketing), GLP-19 (IR cross-`Begin`/`End` batching) |
| 5 — Skill-cast cost | GLP-24 (`$glstats` pass attribution), GLP-25 (IR ring growth), GLP-29 (particle vertex generation) |

Phase 3 (fragment-stage occupancy / early-Z) has no landed work.

Docs are updated alongside: `docs/GPU Skinning/glperf/README.md` catalogs every landed milestone with what was and was not verified, and a staleness pass corrected the IR, GPU-skinning, migration and index documents against the shipped code.

---

## Known-unfixed, documented in the series

- **Particles do not batch.** GLP-19 collapses UI, text and joints (Joints reaches ~26,700 quads in 14 draws) but particles measure **1.0 quads per draw** — `RenderParticles()` changes blend mode about once per particle and each change correctly flushes the pending batch. Grouping by blend state is `GLP-20`; an implementation was attempted, measured to change nothing, and reverted rather than kept on hope.
- **`Joints` CPU cost is not submission.** ~1.1 ms across 9–14 draws is vertex generation. No ticket yet.
- **`Chars` is genuinely GPU-bound** (GPU ~6× CPU). Points at the unconditional `discard` in `BMDMeshShader` defeating early-Z — Phase 3's subject.

---

## Merge state

Merged with `main` at `5cfebed3`; **0 commits behind**. The merge required a non-textual resolution: `c6603879` applies a per-texture terrain UV scale per tile, immediately before a per-tile `glDrawElements` that GLP-16 had removed in favour of bucketed draws. The scale is now carried on the draw bucket and applied once per bucket. That is exact rather than approximate because the scale derives purely from the two bound textures' dimensions, so it is a pure function of the existing bucket key — every tile in a bucket computes an identical value. Owner-verified in-game after the merge: tiles render correctly.

Debug and Release both build clean.


janblade:
Measurements are on debug build.
FPS on release build is too high

image

nitoygo:
Also tested by jkok25 (from discord):  
> works identically, with one exception. The 'Chaotic Disser' skill is untamed. In the old version (Attack speed 600), the 'Chaotic Disser' skill achieves 60 fps - in the current version, 32 fps. With other skills, it's normal - a similar value compared to the previous version.

IntelHD530_5
IntelHD530_3
IntelHD530_4
IntelHD530_1
IntelHD530_6
IntelHD530_2

Notes:
* Chaotic Deseier (Destroyer?) is an existing issue, even before FFP retirement -- will be handled in a separate issue: https://github.com/sven-n/MuMain/issues/559
* The Elf with an equipped Divine Bow releases arrows below the feet -- will be handled in a separate issue: https://github.com/sven-n/MuMain/issues/561

Issue #558, opened 2026-08-12
[Graphics] Support DirectX in Addition to OpenGL


nitoygo:
## Overview

Add support for **DirectX** as an alternative rendering backend alongside the existing **OpenGL** implementation.

The rendering API should be selectable through `config.ini`, allowing the application to choose which graphics backend to initialize and use at runtime without requiring code changes or recompilation.

## Requirements

* Add a DirectX rendering backend alongside the existing OpenGL backend.
* Preserve the existing OpenGL implementation and behavior.
* Introduce a configuration option in `config.ini` to select the rendering backend.
* The application should read the configured backend during initialization.
* Supported values should include at least:

  * `opengl`
  * `directx`
* Define a sensible default backend for backwards compatibility, preferably `opengl`.
* Invalid or unsupported backend values should produce a clear configuration error and/or fall back to the default backend.
* Keep backend-specific code isolated so that rendering logic is not unnecessarily coupled to either OpenGL or DirectX.

## Example Configuration

```ini
[graphics]
renderer=opengl

[graphics.opengl]
CoreProfile=1
```

or:

```ini
[graphics]
renderer=directx

[graphics.directx]
...
```

## Implementation Considerations

* Introduce a common renderer abstraction/interface where appropriate.
* Move renderer-specific initialization, resource management, drawing, and shutdown logic behind the abstraction.
* Avoid scattering checks such as `if renderer == ...` throughout the game/application code.
* Ensure both backends expose the functionality required by the existing rendering pipeline.
* Consider how textures, shaders, buffers, render targets, synchronization, and window/context/device initialization map between OpenGL and DirectX.
* Ensure the selected backend is initialized before any renderer-dependent resources are created.
* Maintain existing OpenGL behavior when no renderer is explicitly configured.

## Acceptance Criteria

* [ ] The application can run using OpenGL through `config.ini`.
* [ ] The application can run using DirectX through `config.ini`.
* [ ] OpenGL remains the default when no renderer is configured.
* [ ] Invalid renderer configuration is handled gracefully.
* [ ] Renderer selection does not require recompiling the application.
* [ ] Common game/application rendering code is independent of the underlying graphics API where practical.
* [ ] Existing OpenGL functionality and performance are not regressed.
* [ ] DirectX-specific code is isolated from the OpenGL implementation.
* [ ] Both renderers can be initialized and shut down cleanly.
* [ ] Documentation/configuration examples are updated to describe the available rendering backends.

[...]

PR #562, opened 2026-08-13
DirectX 11 rendering backend

janblade:
# Add an optional DirectX 11 rendering backend
#558 

Introduces a second rendering backend behind an `RHI` (Render Hardware Interface) abstraction, selectable at runtime. **OpenGL remains the default and is unchanged**; D3D11 is entirely opt-in.

Branch is merged up to current `main` (includes PR #560, the GLP series).

## Why

The FFP retirement and Core Profile migration already pushed all rendering through a small set of shader classes and state wrappers. That made a second backend tractable: the call sites were already funnelled. The motivation is driver stability and frame pacing on modern Windows, access to DirectCompute for cloth physics

This dx-port was made side by side with GL ffp retirement port. Not after the fact.
Because DirectX 11 is a modern API with explicit buffer structures and state management (unlike legacy OpenGL, which allowed loose driver states and implicit operations), adapting to DX patterns revealed several places where the OpenGL pipeline was stalling or doing unnecessary work.

The Direct3D 11 port acted as a strict "code reviewer" for the renderer. Requiring explicit memory flags, interleaved layouts, constant buffers, and indexed triangle topologies in DX forced OpenGL to abandon loose legacy driver patterns and adopt high-throughput GPU streaming.

## Impact on the existing OpenGL path

This is the part worth scrutinising, so stating it plainly:

- **GL rendering behaviour is unchanged.** Every D3D11 addition sits behind `if (g_RenderBackend == RenderBackend::D3D11) { …; return; }` guards placed *before* the existing GL code, or in separate HLSL string literals. No GL branch was rewritten.
- **Default is OpenGL.** Three layers enforce it: `g_RenderBackend`'s static initialiser, the config read default, and a parse function that only returns D3D11 on an explicit match. An unrecognised value falls back to GL rather than silently selecting the newer backend.
- **Two changes do affect GL users**, and neither is silent:
  - `config.ini` gains a `[graphics]` schema (below) — auto-migrated, no manual edit.
  - The crash report's `Direct-X 7.0` line is gone, replaced by an accurate `Renderer :` line. That line came from a DirectDraw-era probe that caps at 9.0, knows nothing of D3D10/11/12, and dead-ends at 7.0 on modern Windows because `d3d8.dll` no longer ships — so it read as "this client renders with Direct3D 7" regardless of reality.

## Architecture

`RHI` dispatches on `g_RenderBackend` through a shim (`RHI.cpp`) — no virtuals, branch on a value read once at startup. `RHI_GL.cpp` wraps the existing GL calls; `RHI_D3D11.cpp` is new.

Each of the four shaders (`Passthrough`, `Terrain`, `PlanarShadow`, `BMDMesh`) carries a hand-ported HLSL twin alongside its GLSL source, compiled at runtime via `D3DCompile`. Constant buffers mirror the GLSL `std140` layouts byte-for-byte, `static_assert`-ed.

## What's included

| Area | Notes |
|---|---|
| Device / swapchain / resize | `RHI_D3D11.cpp` |
| All four shaders in HLSL | Same `Create`/`Destroy`/`Bind` shape as the GL path |
| Blend, depth, cull, rasterizer state | Eight blend combos as pipeline state objects |
| Textures, vertex/index buffers, cbuffers | Behind the existing `RHI` calls |
| GPU compute cloth | `ClothComputeShader` — DirectCompute solver for capes/wings. **Off by default**, opt-in via `GpuCloth=1` |
| Clip-space handling | Shared projection builder; GL `[-1,1]` vs D3D `[0,1]` |

## Config change

`[Render]` becomes a selector plus per-backend subsections, so a key can't be ambiguous about which backend it applies to:

```ini
[graphics]
renderer=opengl           ; or "directx"

[graphics.opengl]
CoreProfile=1
MaxGLVersion=

[graphics.directx]
D3D11DebugLayer=0
GpuCloth=0
```

Existing configs migrate on first launch and `[Render]` is deleted. `Backend=D3D11` → `renderer=directx`, so anyone already using the old key keeps their choice. Keys never explicitly set aren't written, so they keep tracking compiled-in defaults rather than being frozen at current values.

## Known caveats

Being upfront rather than discovering these in review:

- **Maturity.** D3D11 renders the world, characters, UI and effects and reaches live gameplay, but it has had a fraction of GL's playtime. Expect rough edges; `renderer=opengl` is the fallback for isolating anything suspect.
- **Maintenance cost.** The GLSL and HLSL twins must move in lockstep, and they live in different string literals in the *same* file — so a change to one merges cleanly against the other with no conflict marker, and the build stays green while the backend is broken. This bit me three times bringing the branch up to current `main` (bone-palette layout, a deferred-draw-batch invariant, and shared-buffer layout drift). `docs/DirectX11/gotchas-and-patterns.md` §5 documents the failure mode and classifies which GL changes need a D3D11 counterpart.
- **GPU cloth is opt-in and still soaking** — CPU remains the real path for every cloth instance.
- **FSR1/FSR2/DLSS are scoped only**, not implemented.

## Verification

- Debug + Release build clean.
- Confirmed in-game under D3D11: world, terrain, characters, objects, UI text, and live gameplay.
- Config migration verified at runtime against a real legacy file, idempotent across launches, both `renderer` values selecting the correct backend.
- Particle blend-state fix **confirmed in-game** in a particle-heavy scene. (This was listed as unverified when the PR was opened.) The black-rectangle artifact on smoke/dust is resolved; it needed two parts — four blend-mode wrappers that had no D3D11 branch at all and silently fell through to no-op GL calls, plus flush ordering, since `RHI::SetBlendMode` binds immediately while the immediate-mode renderer defers its draw.
- ⚠️ D3D11 has **no performance benchmarks** — no claim of parity with or improvement over GL is being made here.

## Docs

New `docs/DirectX11/` covering backend architecture, the compute cloth pipeline, the DXP milestone catalog, and a gotchas file (clip-space, Y-flip, cbuffer padding, state-guard rules, cross-backend lockstep).


sven-n:
I read the Why paragraph, but... why?!
I think there are so many other things to improve that adding another renderer isn't worth the effort.

nitoygo:
Hello @sven-n, I understand the concern about stabilizing the new OpenGL renderer first, but I don't think adding DirectX necessarily has to wait until OpenGL is considered fully stable. They can actually be developed and debugged side by side, with issues in the renderer abstraction becoming more apparent when both backends exercise the same code. 

DirectX is being proposed as a practical compatibility path for older Windows PCs and integrated/low-end GPUs where the Direct3D driver may perform better or be more reliable. If the additional backend can be added without significantly slowing down OpenGL stabilization, I don't think it's wasted effort.

As jkok25 tested the branch, he noticed that DirectX is actually better for his setup than the OpenGL one:

> jkok25 — 22:45
I tested this with DirectX:
https://github.com/janblade/MuMain/tree/dx-only-port
For an Intel HD 530, this is the best option. 😍 
With two windows (2 acc), it's playable at a fixed 40 fps setting ($fps 40) - spot: 11 mobs. This is an excellent achievement.
Similar machines are often used these days. My system is Windows 11 - DirectX 12, but it will work the same on Windows 10 - DirectX 11.

https://github.com/user-attachments/assets/4f498497-2fa2-40a5-ac96-28e8e65ca96e

> For comparison, OpenGL - HD 530.
(fps drops to 20):

https://github.com/user-attachments/assets/121802fd-0cd8-40c5-9d42-8850d508f053

:

sven-n:
Understood... however, I think there is still a lot to gain with OpenGL, because the code isn't really making use of shaders. Yes, it uses them, but the CPU still does a lot of the heavy lifting.
Did you try release build, btw? Debug is a lot slower, even on my fast machine.

janblade:
Actual flags in CLIENT/out/build/windows-x64/CMakeCache.txt:


CMAKE_CXX_FLAGS_DEBUG   = /Ob0 /Od /RTC1
CMAKE_CXX_FLAGS_RELEASE = /O2  /Ob2 /DNDEBUG
Four things stack up, roughly in order of how much they cost this codebase:

1. /Od /Ob0 — no optimization, no inlining at all.
This is the big one for MU. The hot paths are tiny functions called millions of times per frame: vector/matrix helpers, IR::Vertex3f(), bone-palette math, terrain tile loops. In Release each of those disappears into the caller. In Debug every one is a real call with a stack frame, and locals live in memory instead of registers. /Ob0 also disables inlining you explicitly asked for — even inline and header-defined one-liners are emitted as calls.

2. _ITERATOR_DEBUG_LEVEL=2 — checked iterators.
[CLIENT/src/CMakeLists.txt](vscode-webview://0tqtbqu1eib69kbeuf4rq7h673ts1joufqeso53bodutt1einh5g/CLIENT/src/CMakeLists.txt) sets no _ITERATOR_DEBUG_LEVEL / _HAS_ITERATOR_DEBUGGING override (grep-confirmed again just now), so MSVC's Debug default applies. Every std::vector op gets bounds validation plus iterator-lifetime bookkeeping. ImmediateRenderer.cpp's Vertex3f() calls push_back 6+ times per quad — that single path runs 10–50× slower in Debug. It's why 2,975 particles measured 5.57 ms (~1.87 µs each) for what is a 3×4 matrix multiply and ten appends.

3. /MDd debug CRT heap.
Every allocation gets guard bytes, fill patterns (0xCD/0xDD), and a linked-list of live blocks. Anything allocating per-frame pays a multiple of Release cost.

4. /RTC1 runtime checks.
Injects uninitialized-local detection and stack-frame poison/verify code around effectively every function.


The consequence that keeps biting this project: absolute CPU-ms from a Debug capture are inflated by an unknown factor and mean nothing about shipped performance. Debug-vs-Debug before/after comparisons stay valid, and GL/draw/BufUpd counts are build-independent — but the whole GLP-19/24/25/26/29 Phase 5 thread was characterised on Debug numbers before that was caught. If you want profiling with symbols, use RelWithDebInfo (/O2 /Zi), not Debug.

janblade:
We already measured using $glstats. CPU ms below 1ms for every item.
Except with some unoptimized skill animations.

If release build can already reach, 300 to 400 + FPS, not much to gain on CPU side.
But yes, need to optimize some skills. 
No skills should ever half the Frame rate on continuous cast.

If we fix that, both GL and DX path will benefit. not Just GL.

Mosch0512:
i am strigtly against it. it will introduce an complete extra tec stack to maintain and the code is and will be harder to maintain due to bad structure getting even worse ... i would just fix the performance issues with opengl like shaders mentioned ... also i do not like if your respned with ai to us ...who is prooving a point an ai that can haluzinate or an human having a strong opinion that could be taked about?

Mosch0512:
the only reason i see if performance ... as you stated for some older gpus like intels integrated ones opengl might performe badly but i dont think it will perform good on an raspberry pi aswell so should we optimize it for that too? where do we start and where do we stop? we are only introducing more mess to a insanly messy code that is going to bite use sooner or later ... i think

janblade:
Retiring FFP was the mandatory prerequisite that made further CPU/GPU decoupling and engine refactoring possible. Trying to jump directly from a 2001 fixed-function renderer to zero CPU-GPU communication without eliminating FFP first would be practically impossible.

Detailed Analysis of the Feedback Claims
1. "The only thing the CPU should communicate to the GPU is a timestamp..."
Where this is true: In simple WebGL demos, CAD viewers, or modern engines with pure GPU-driven pipelines (where animation tables, scene graphs, particle state, and indirect draw buffers reside entirely in VRAM), the CPU's per-frame job is indeed reduced to updating frame timers and dispatching compute passes.
Where this falls short for MU Online: MU Online is a 2000s C++ MMORPG codebase where game logic, combat hitboxes, click-picking, particle attachments (CreateParticle, CreateJoint), and character states are deeply coupled with CPU data structures (CHARACTER*, OBJECT*, stack matrices).
Expecting a legacy 300k+ line C++ codebase to leap directly to modern GPU compute/indirect rendering in a single step—without first eliminating legacy glBegin/glEnd and glMatrixMode calls—is an unrealistic migration path.
2. "The CPU does too much, the GPU does too little, and CPU-GPU communication is the bottleneck"
This observation is accurate regarding the legacy baseline:
Historically, MU Online's engine executed vertex skinning per-vertex on the CPU (BMD::Transform()), calculated planar shadow ground skews on the CPU, and streamed dynamic vertex data via immediate-mode calls (glBegin(GL_QUADS) / IR::Begin).
Per-mesh glGetFloatv(GL_MODELVIEW_MATRIX) readbacks were a measured runtime bottleneck (e.g., dropping Lorencia frame rates from ~110fps to ~63fps during initial raw VBO ports before matrix caching was introduced).
However, retiring FFP is what allowed offloading work to the GPU:
GPU Bone Skinning: Skeletons (up to 200 bones per character) now skin directly in vertex shaders (u_Bones[200]).
GPU Planar Shadows: Ground skew projection moved into single-pass vertex shaders (CPlanarShadowShader).
Item Specular Shader: Tiered armor shine (+7 through +15 gear) offloaded to 4-variant GLSL/HLSL shaders.
Terrain Tile Bucketing: Reduced terrain submission overhead from ~1,226 per-tile draw calls to ~48 bucket calls.
Why FFP Retirement Made Refactoring EASIER (Not Harder)
Retiring FFP was not an end in itself—it was the foundational cleanup required to make modern refactoring possible:

Implicit Driver State $\rightarrow$ Explicit Data Contracts:

Under FFP, rendering relied on hidden driver state machines (glGetFloatv(GL_MODELVIEW_MATRIX), glColor4ub, ambient lighting flags, fixed texture units). You could not decouple CPU and GPU logic because state was scattered across thousands of legacy draw calls.
FFP retirement forced clean data interfaces: GlobalUBO / BoneUBO (constant buffers), interleaved vertex structures ([Pos | UV | Color]), and explicit shader uniforms.
RHI Abstraction Layer & Portability:

FFP retirement enabled building the RHI (Render Hardware Interface) layer (RHI_GL / RHI_D3D11).
Because drawing now goes through RHI abstractions (RHI::AppendBuffer, UploadBones, RHI::DrawIndexed), any future optimization implemented above the RHI line (such as draw batching or GPU instancing) automatically benefits both OpenGL and Direct3D 11 simultaneously.
Safe, Incremental Migration Path:

Large C++ engine modernizations fail when attempted as all-at-once rewrites.
Moving off FFP established a clean, stable baseline where individual subsystems (e.g., UI rendering, terrain, particle systems, dynamic mesh buffers) can now be optimized for GPU-only data flow one at a time without breaking game logic or cross-platform compatibility.
Conclusion
Is the feedback true? The observation that CPU-GPU bandwidth is a key performance bottleneck is true. The suggestion that FFP retirement was useless or should have been deferred until after a total engine refactor is false.
Is it easier to refactor now? Yes, significantly. With FFP retired and explicit data structures (UBOs, VAOs/VBOs, clean shaders, RHI dispatch) in place, the engine is finally in a state where further GPU offloading (e.g., GPU instancing, compute particles, indirect draws) can actually be implemented cleanly.

janblade:
And to be fair with DX11 and AI, by exploring DX11, we were able to copy modern DX ways of doing things back to GL. 

With this baseline of core profile + dx11. 
it is now easier and safe to refactor and be compliant with both.

instead of refactoring first then the refactored code breaks
core profile and dx11 requirements. 
then you have to refactor/redesign again.

janblade:
This again from AI.. but tell me it's not true

# Direct3D 11 Inspired OpenGL Pipeline Optimizations

Because DirectX 11 is a modern API with explicit buffer structures and state management (unlike legacy OpenGL, which allowed loose driver states and implicit operations), adapting to DX patterns revealed several places where the OpenGL pipeline was stalling or doing unnecessary work.

Here are the 4 primary examples from this codebase:

### 1. Dynamic Buffer Streaming (`AppendBuffer` & `WRITE_NO_OVERWRITE`)
* **How DirectX does it better**: D3D11's `ID3D11DeviceContext::Map` API forces explicit buffer streaming flags—specifically `D3D11_MAP_WRITE_DISCARD` (re-allocate buffer) versus `D3D11_MAP_WRITE_NO_OVERWRITE` (append to existing ring buffer without stalling the GPU pipeline).
* **How OpenGL benefited**: Legacy OpenGL was re-uploading dynamic buffers via repeated `glBufferData` calls, causing driver-level pipeline stalls (the GPU had to wait for the CPU to finish writing). Adopting D3D11's ring-buffer pattern in [`RHI_GL.cpp`](file:///d:/DEV/MUONLINE%20WORKSPACE%20UPSTREAM/CLIENT/src/source/Render/RHI/RHI_GL.cpp) (`RHI::AppendBuffer`) eliminated CPU-GPU sync stalls for UI, particles, and immediate-mode draws.

### 2. Interleaved Vertex Buffers (3x Uploads $\rightarrow$ 1x Upload)
* **How DirectX does it better**: D3D11 requires vertex data to be bound via structured input layouts with explicit byte strides per input stream.
* **How OpenGL benefited**: MU's legacy OpenGL renderer uploaded positions, texture coordinates, and vertex colors as three separate VBO streams per mesh. Following D3D11's layout structure, OpenGL was refactored to use a single interleaved buffer (`[vec3 Pos | vec2 UV | vec4 Color]`, 36-byte stride). This cut GPU upload bus calls from 3 per mesh down to 1 per mesh (see [`architecture_overview.md`](file:///d:/DEV/MUONLINE%20WORKSPACE%20UPSTREAM/.ai-os/memory/semantic/knowledge/architecture_overview.md)).

### 3. Immediate Primitive Decomposition (Quads/Fans $\rightarrow$ Triangle Lists)
* **How DirectX does it better**: Direct3D 11 completely omitted support for legacy `D3DPT_QUADLIST` or `GL_QUADS` primitive topologies; all quad and fan geometry must be indexed triangle lists (`D3D11_PRIMITIVE_TOPOLOGY_TRIANGLELIST`).
* **How OpenGL benefited**: `ImmediateRenderer` ([`ImmediateRenderer.cpp`](file:///d:/DEV/MUONLINE%20WORKSPACE%20UPSTREAM/CLIENT/src/source/Render/Core/ImmediateRenderer.cpp)) was refactored to decompose `GL_QUADS` ($0,1,2 + 0,2,3$) and `GL_TRIANGLE_FAN` primitives into indexed triangles CPU-side. This eliminated legacy `glBegin(GL_QUADS)` fixed-function calls in OpenGL and allowed modern `glDrawElements` execution across both backends.

### 4. Uniform Buffers vs. Individual `glUniform*` Calls (`GlobalUBO` / `BoneUBO`)
* **How DirectX does it better**: D3D11 handles shader constants strictly through Constant Buffers (`cbuffer` bound to registers `b0`, `b1`, etc.) containing 16-byte packed structs (`std140` rules).
* **How OpenGL benefited**: Legacy OpenGL code queried and set individual uniform locations one-by-one (`glUniform1i`, `glUniformMatrix4fv`). Adopting D3D11's constant buffer design led to `GlobalUBO` and `BoneUBO` ([`BoneUBO.cpp`](file:///d:/DEV/MUONLINE%20WORKSPACE%20UPSTREAM/CLIENT/src/source/Render/Core/BoneUBO.cpp)), allowing OpenGL to upload entire matrix palettes (200 bone matrices per character) in a single block upload instead of dozens of individual uniform calls.

---

### Summary
The Direct3D 11 port acted as a strict "code reviewer" for the renderer. Requiring explicit memory flags, interleaved layouts, constant buffers, and indexed triangle topologies in DX forced OpenGL to abandon loose legacy driver patterns and adopt high-throughput GPU streaming.

janblade:
> I read the Why paragraph, but... why?! I think there are so many other things to improve that adding another renderer isn't worth the effort.

updated

sven-n:
> **Transparency note:** this review was written by Claude (an AI assistant), run by me against the actual branch at `ae6bcc0`, at my request. Given @Mosch0512's fair point earlier in this thread about AI-generated arguments, I'm labelling it rather than passing it off as hand-written. Every claim below is a code reference you can check yourself — please do. The conclusions are ones I agree with, and the call on the PR is mine.

Thanks for the detailed writeup and the honest "Known caveats" section, @janblade — the FFP retirement and GLP series were real work and I don't want any of the below to read as dismissing it. But I've now gone through the branch rather than the description, and I don't think this should merge as-is.

## 1. ~28% of the diff is code nothing calls

`ClothComputeShader.cpp/.h` is 1,826 lines. Its only call site outside itself is `RunSelfTestAndLog()` in `Winmain.cpp:1802`. The flag that supposedly enables it:

```
RenderConfig.cpp:12    bool g_GpuClothEnabled = false;
RenderConfig.cpp:160   g_GpuClothEnabled = (gpuCloth != 0);
```

Those are the only two references in the tree — nothing reads it. `CPhysicsCloth` is not modified by this PR at all.

So the PR body's "GPU compute cloth — DirectCompute solver for capes/wings. **Off by default**, opt-in via `GpuCloth=1`" isn't accurate: setting `GpuCloth=1` does nothing. It's a self-test harness plus a solver with no caller. That needs to come out of the PR regardless of what we decide about D3D11 itself.

## 2. The RHI abstraction isn't holding a line

- **93** `g_RenderBackend` branch points in the diff — 25 in `ZzzOpenglUtil.cpp` alone. The backend split lives *inside* call sites rather than at a boundary.
- `RHI::GetCaps()` in `RHI.cpp` returns `RHI_GL_Impl::GetCaps()` unconditionally, regardless of backend. `ErrorReport.cpp:258` reads it — so under D3D11 crash reports print GL caps from a device that was never initialised. Ironic given the PR replaces the bogus `Direct-X 7.0` line for accuracy reasons.
- `RHI.h` exposes `GetD3D11Device()`, `GetD3D11DeviceContext()` and `RegisterExternalD3D11VertexBuffer()` — backend-specific escape hatches on the backend-agnostic interface, declared in the `RHI_GL_Impl` namespace as well.

## 3. No GL context under D3D11, but 536 raw `gl*` calls remain

`GLColorIntercept.h` states it directly: *"under Backend=D3D11 there is no GL context at all."* Meanwhile 536 raw `gl*` call sites survive across 29 files, ~25 of them outside the RHI backends. Under D3D11 each is one of two things:

- **Silent no-op** — GL 1.1 entry points from `opengl32.dll`, which fail quietly with no current context. `FrustumRenderer.cpp` (debug wireframe) and `GMBattleCastle.cpp:581` (map fog) are ungated and simply stop working.
- **Access violation** — anything resolved through `SDL_GL_GetProcAddress`. Which is exactly why whole subsystems had to be gated; `LoginScene.cpp`'s own comment records it as *"the difference between 'black world, login screen renders' and 'access violation before the first frame completes'."*

The gating is also inconsistent in a way that undercuts the "confirmed in live gameplay" claim. `LoginScene` and `CharacterScene` gate out `RenderBlurs` / `RenderJoints` / `RenderEffects` / `RenderPets` / `RenderBoids` / `RenderObjects_AfterCharacter` / `CheckSprites` under D3D11 — but **`MainScene.cpp` is untouched by this PR**, so those same functions run ungated during actual gameplay. Either the gates are unnecessary, or gameplay is working by luck. It can't be both, and I'd want to know which before shipping.

## 4. The lockstep cost

You're upfront that the GLSL and HLSL twins live in separate string literals in the same file, so a one-sided change merges clean with no conflict marker, the build stays green, and the backend is silently broken — and that this caught you three times just rebasing onto `main`. I appreciate the honesty, but that's the part I can't get past: it's not a documentation problem, and nothing in the build or CI can detect it. `gotchas-and-patterns.md` §5 documents the failure mode; it can't prevent it.

The inline comments in the diff read as a log of exactly that: D3D11-only viewport Y-flip leaving a black bar, `RenderColor()` silently resetting blend state so every text background box rendered opaque, `IR::Flush()` on the wrong side of a branch leaking a batch into the next frame, character-select unusable because selection was wrongly gated. All found by playing the game, none catchable by the build. That's the ongoing cost of a second backend in this codebase, and it lands on whoever touches the renderer next.

## On "DX made the GL path better"

I think this is true, and it's the part of your argument I accept. But that value is **already banked** — interleaved vertex buffers, the UBO work, triangle-list decomposition, ring-buffer streaming are all merged in `main` via the GLP series. It's a strong argument for having done the exploration. It isn't an argument for merging and then maintaining the backend forever.

## On the HD 530 result

@nitoygo — thank you for bringing the numbers, and thanks to jkok25 for testing. That result is the one thing here I genuinely weigh, and low-end integrated GPUs are a real constituency for this client. But the PR itself says "no performance benchmarks — no claim of parity with or improvement over GL is being made," which leaves one anecdote against a permanent second tech stack. If integrated Intel is the actual motivation, I'd much rather spend the effort finding out *why* GL is slow there — it may well be a specific stall or driver path, and a fix in GL helps every user rather than only those who edit config.ini.

## Where I'd like to land

1. **Split out the `ErrorReport.cpp` change now.** Removing the DirectDraw-era probe that reports `Direct-X 7.0` on every modern machine (−257 lines) is a clean win with no D3D11 dependency. Send it as its own PR and I'll merge it.
2. **Drop `ClothComputeShader` and its two config keys** from this PR. Unreachable code shouldn't land.
3. **If D3D11 is to stay on the table**, the precondition for me is making the lockstep failure mechanically detectable: shaders in separate per-backend files so a one-sided edit conflicts or fails to build, and the backend split lifted to a real boundary instead of 93 inline branches. Without that, every future renderer change costs double and half of it breaks silently.
4. **The `[graphics]` config restructure** is reasonable on its own merits and could go in separately if you want it.

I know this isn't the answer you were hoping for after the amount of work in here. The GL improvements this exploration produced are genuinely valuable and they're staying. I'd just rather put the next stretch of effort into making one renderer fast than into keeping two correct.

---

PR #564, opened 2026-08-14
perf(render): group particle draws by texture (opt-in)

sven-n:
## Summary

Particles are walked in slot order — allocation order — so textures interleave. `BindTexture()`'s dirty check flushes the `IR::` batch whenever the texture changes, which is nearly every sprite, so each particle becomes its own draw call.

This looks like the reason **GLP-19 measured nothing for particles**: the cross-`Begin`/`End` batching works fine, it just never gets two consecutive sprites to merge. Grouping draws by texture is what gives it something to merge.

Two commits:

1. **`refactor`** — extract `RenderParticle()` from `RenderParticles()`' ~390-line loop body, so the loop can choose an order without touching the draw code. No behaviour change; the 361 extracted lines are identical apart from indentation and one rename (`i` → `particleIndex`, for the two cases that derive their look from the particle's slot). Verified by line-by-line comparison against the previous revision.
2. **`perf`** — the ordering pass itself, behind `[Render] SortParticleDraws`, **default off**.

### Why the reordering is constrained

Reordering draws is only sound where the pixels cannot depend on the order, so nothing is reordered freely. A particle joins a reorderable run only if all three hold:

- **Its blend mode is additive** (`GL_ONE, GL_ONE`). Additive accumulation is commutative, so those sprites permute freely. The alpha-tested branch is not, and neither are the subtractive/alpha-blended cases.
- **Its draw case does not pick its own blend mode.** 14 effect types call `EnableAlphaBlendMinus()`/`EnableAlphaBlend3()` inside their case instead of keeping the additive default.
- **It does not toggle the depth test.** The two depth toggles are *sticky* — nothing restores them, so once one fires every particle drawn afterwards in the frame inherits the change. This was the subtlest constraint and the one most likely to be missed.

Everything else keeps its exact position and acts as a **barrier no draw is moved across**. That makes the classifier free to be as conservative as it likes: at the limit this degrades to a no-op rather than to a wrong picture. Grouping is `stable_sort`, so same-texture particles keep their relative order and the output is *identical*, not merely equivalent — which is what makes an A/B comparison meaningful. The draw path calls no `rand()`, so it stays deterministic too.

The barrier list was derived by sweeping every case body in `RenderParticle()` for blend and depth calls rather than from memory, and every listed label was checked to exist in that switch. **It has to be kept in step with that function** — a missing entry is a real ordering bug.

### Verification

- Full Linux client builds clean (Release).
- `ctest`: **30/30 pass**, 10 new — barrier containment (nothing crosses one), stability within a texture, per-stretch independence, barriers at the ends and adjacent barriers, degenerate inputs, and that grouping is a permutation (no draw dropped or duplicated).

### Not verified — please read before flipping it on

- **Whether real effects still look right.** A build cannot answer this and I have no GPU or game assets here. That is why it defaults to off.
- **Whether it actually wins anything.** Several of the most common effect types — `BITMAP_SMOKE`, `BITMAP_FIRE`, `BITMAP_CLOUD`, `BITMAP_SPARK`, `BITMAP_LIGHT + 2` — are on the barrier list, so the reorderable population may be smaller than it first looks and the win correspondingly modest. A `$glstats` capture with the switch off vs on (draw calls and `IR` flushes during a Chaotic Diseier cast) is the honest test, and it should happen before this is considered for default-on.

If the measurement shows the barrier list is what is limiting the win, the next step is not to loosen it but to make the barrier cases cheaper — atlasing the common effect sprites, or moving the sticky depth toggles to explicit per-particle state.

## Checklist

---
_Generated by [Claude Code](https://claude.ai/code/session_01HLvvZa4qDXHWG1LcE8Z9ke)_


sven-n:
Correcting the premise of this PR — I found a prior measurement after opening it that I should have found first.

`docs/GPU Skinning/gotchas-and-patterns.md` §4.2 records a hand-tagged flush-site capture: **`blend: 2,940` against 2,918 particles — roughly one blend change per particle.** So the dominant batch-breaker in the particle path is measured to be *blend*, not texture.

That does not make this change wrong, but it narrows what it can be expected to do. The reorderable set here is additive-only, so blend is constant *within* a reorderable run and grouping by texture removes the remaining texture breaks inside it. What it cannot do is fix the blend churn itself: every alpha-tested particle ends a run, so if additive and alpha-tested particles alternate, runs are length ~1 and this does nothing at all. The win depends entirely on how much additive particles already cluster in slot order, which nobody has measured.

The larger prize would be grouping by `(blend class, texture)` — but moving particles across blend classes is not order-independent, so that needs a different argument than the one this PR rests on, not just a wider sort key.

Recommend landing #566 (IR batch-break attribution) first and reading the `Particles` row's `Vtx/Draw` with `SortParticleDraws` off vs on. If `Vtx/Draw` stays near 6 with it on, this PR is not the right lever and should be closed rather than merged and left switched off.


---

sven-n:
It seems to have a positive impact. I just tested debug build yet and did no visual inspection, but it looks promising.

PR #566, opened 2026-08-15
feat(profiling): attribute IR batch breaks by cause in $glstats

sven-n:
## Summary

`$glstats` counts draw calls but says nothing about *why* consecutive `IR::` quads failed to merge. That gap has already cost this series real time.

`gotchas-and-patterns.md` §4.2 records it: instrumenting `Begin()`'s merge decision reported **`key mismatches: 0`** on a frame with **2,918 unbatched particle draws** — which reads as "render state never changes" and is the exact opposite of the truth, because state writers flush *before* mutating, so the batch is already gone by the time `Begin()` compares anything. A correct earlier finding (that `RenderParticles()` churns blend mode per particle) was **retracted** on the strength of that reading. The real answer only appeared after someone tagged the flush call sites by hand and got `blend: 2,940` against 2,918 particles.

This makes that tagging permanent.

### What's counted

Both inside `Flush()`, past its pending-batch check — most `Flush()` calls find nothing pending, and counting those would drown the signal:

- **`IRDraws` / `IRVertices`** — `Vtx/Draw` is the headline. An IR quad is 6 vertices after triangle decomposition, so a ratio near 6 means every sprite is still its own draw call and GLP-19's batching is merging nothing. Well above 6 means runs are merging.
- **`IRBreak{Texture,Blend,Depth,Program,Uniform,Matrix,Draw,Other}`** — tagged at all 26 flush sites. This is where a batching fix has to aim. `Draw` is its own bucket because "another draw call cut in between two IR batches" is a distinct and actionable cause from a state change.

Both are tags, never GL calls: they don't bump `GLCalls`, and `IRDraws` is a subset of `DrawCalls` that `RHI::Draw` already counted.

The overlay gains a frame-total `IR` line, the `IRbreak` breakdown, and a per-pass `Vtx/Draw` row for `Sprites`/`Particles`/`Joints`/`UI` — the first three for the same reason GLP-24 split them out of `Other`: they were the frame's largest draw-call producers while being invisible. Everything stays behind `g_CountersEnabled`, so `$glstats off` is unchanged.

Also updates the §4.2 "Related" bullet, which said `$glstats` has no blend-mode counter. It does now.

### Verification

- Full Linux client builds clean (Release); 20/20 existing tests pass.
- The counters are only meaningful once read in-game — **no capture has been taken**. That is the point of the PR, not a gap in it.

### What to capture with it

Idle / crowded / mid-Chaotic-Diseier-cast, Release, on the Intel HD 530. Specifically:
- `Particles` row `Vtx/Draw` — if it sits near 6, batching is doing nothing there.
- The `IRbreak` split — §4.2's hand measurement says **blend**, not texture, is the dominant particle batch-breaker.

That last point matters for #564 (particle texture grouping), and I've flagged it there: if blend still churns once per particle, grouping by texture only helps to the extent additive particles already sit next to each other. This PR is how you find out, and it's worth landing first.

## Checklist

---
_Generated by [Claude Code](https://claude.ai/code/session_01HLvvZa4qDXHWG1LcE8Z9ke)_


sven-n:
It's very helpful IMHO. When casting specific skills, you can see what causes fps drops. In case of chaotic desire it's a huge amount of particles.

PR #570, opened 2026-08-18
[WIP]Add a native macOS client along the Linux port seam

afrokick:
## Summary
- Native macOS build next to the Linux port: CMake `APPLE` path, `osx-arm64`/`osx-x64` Native AOT dylib, Unix shims, and `src/source/App/Platform/macOS/main.cpp`.
- Docs under `docs/build/macos/` plus CI in `.github/workflows/macos-build.yml`. `build-macos/` is gitignored like the other host build dirs.
- Also enables frustum culling on the login scene.

## Test plan
- [ ] `cmake -S . -B build-macos -G Ninja -DCMAKE_BUILD_TYPE=Release -DENABLE_EDITOR=ON` with Homebrew `glew`/`jpeg-turbo` prefixes, then `cmake --build build-macos`
- [ ] Run `./Main` from `build-macos/src` and confirm the client starts
- [ ] Confirm `MUnique.Client.Library.dylib` sits next to `Main` and is a Mach-O dylib
- [ ] Check the login scene still renders and that CI `macos-build.yml` is green

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/workflows/macos-build.yml
+++ b/.github/workflows/macos-build.yml
@@ -1,0 +1,78 @@
+name: macOS Build
+
+# Builds the full native macOS client: the engine, the SDL/OpenGL runtime,
+# and the osx-arm64 / osx-x64 Native AOT network library
+# (MUnique.Client.Library.dylib), for both editor configurations.
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+
+jobs:
+  build-macos:
+    name: build-macos (editor-${{ matrix.editor }})
+    runs-on: macos-latest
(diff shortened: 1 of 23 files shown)


nitoygo:
Thanks for the work on the macOS platform! Since the project has also moved to SDL3, it looks like some of the macOS-specific changes may already have been addressed as part of that migration. Could you take a look and see whether your changes are still applicable, or if they’ve already been resolved by the [SDL3 work](https://github.com/sven-n/MuMain/pull/572)?

PR #571, opened 2026-08-18
feat(benchmark): measure graphics changes with $bench

sven-n:
## Summary

Adds a segmented benchmark to the client, so a graphics change can be judged against a measurement instead of an impression.

`$bench run` measures a list of named segments, each on its own, and writes a run directory under `bench/runs/`. `$bench run fx.*` or `#particles` runs a subset for faster iteration; `$bench quick` is one short repeat for finding out whether a change did anything at all. Usage is documented in [`docs/benchmark.md`](docs/benchmark.md).

**Why segments.** Splitting a run is what lets the result say *where* a change had an effect, rather than only whether the frame got faster. This first catalog varies the effect surfaces of whatever scene you are standing in, using the `$effects` toggles that already exist, so it needs no scripted content:

| Segment | Isolates |
|---|---|
| `scene.full` | Nothing disabled — the reference, and the drift control |
| `fx.all.off` | Every effect surface at once, exactly as `$effects off` does it |
| `fx.sprites.off` / `fx.particles.off` / `fx.joints.off` | The three IR per-quad paths |
| `fx.skillmodels.off` | Skill effect models, impact effects still rendering |
| `fx.boids.off` | Ambient wildlife — an always-on town cost |
| `fx.wingshadow.off` / `fx.winglayers.off` | The extra wing shadow draw, the glow overlay passes |

**What it exports.** `run.json` is canonical and schema-versioned; `report.md`, `frames.csv` and `passes.csv` are views rendered from it at fixed precision so two reports diff cleanly. The report header carries the build configuration, GPU and driver, the GL context version the chain actually settled on, every `RHI::Caps` flag, the resolution and the vsync state — the facts that decide which code path is running, which have so far had to be read out of a boot log by hand.

**What it does to avoid lying.**

- A repeat is a fixed frame count, never a fixed duration — a faster build has to render the same frames, not more of them.
- Each repeat discards a warmup window (shader compiles, texture uploads, streaming-ring growth). A frame over half a second discards its whole repeat rather than averaging an alt-tab in.
- The baseline is measured first and again last whatever the selection was; the difference is drift, and if it is large the report says nothing measured in between is trustworthy.
- The summary carries the 1% low, the percentile curve, the pacing metrics and the **repeat spread** — the noise band a delta has to clear before it means anything. This is the number GLP-29 did not have.
- Findings are rule-based and named, never prose. Two matter most: the **attribution check** (pass CPU ms against frame time — the automatic version of the blind spot GLP-24 found after it had already invalidated a phase of measurements), and **`segment-inert`** (a segment that submitted the same draw calls as the baseline disabled nothing here, so its timing is not evidence that what it targets is cheap).

**Two supporting changes.** `$glstats` used `FrameProfiler::g_CountersEnabled` as both "count things" and "draw the overlay", so a run turning the counters on would have switched a several-hundred-draw-call text overlay on inside its own measurement — the overlay now has its own flag, behaviour otherwise unchanged. The compile-time build info the `$details` overlay printed inline moved to `Core/Utilities/BuildInfo.h`, since the exports need the same facts.

Not in this PR, and worth doing next: scripted scenes (spawned characters, camera paths, weather), per-segment screenshots with an A/B image diff, and a `benchdiff` tool that compares two runs with a compatibility gate on the manifest hash.

## Verification

- **Run in the client**, on an RX 6800 in Lorencia, and that run found a real bug ([#issuecomment-5461948274](https://github.com/sven-n/MuMain/pull/571#issuecomment-5461948274)): `fx.all.off` set only `SetDisableEffects` and not the options switch `$effects off` also flips, so it left sprites, particles and joints rendering — 162 particle draws/frame against the baseline's 171. Fixed in 22bb352, with two tests covering it.
- The same run confirms the rest of the mechanism works: `fx.particles.off` took the Particles pass from 170.6 draws / 0.61 ms to 0.00 across 900 frames, and `fx.joints.off` reached 99.4 FPS against the baseline's 92.5.
- That run also had vsync on, which pinned every median between 9.89 and 10.63 ms and made the segments look alike. The report flagged it afterwards; it is now said in chat as the run starts, while there is still time to `$vsync off`. A clean uncapped re-run is still outstanding.
- 36 doctest cases / 167 assertions over the pure layer (stats, glob selection, findings rules, and the exact bytes of all four exports), passing in CI on Linux and MinGW.
- Full client build green on all six CI jobs: Linux and MinGW x86_64/i686, editor on and off.

## Checklist

- [x] I have built the project locally (see [`docs/build/README.md`](docs/build/README.md)) — full builds via CI on all six configurations; unit tests and per-file syntax checks locally.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TnwayPjwJEQG17DLXNyYzQ)_

### DIFF
--- a/docs/benchmark.md
+++ b/docs/benchmark.md
@@ -1,0 +1,133 @@
+# Client Benchmark
+
+The client can measure itself. `$bench` runs a fixed list of scenarios, records every frame,
+and writes a report you can put next to another one — so a graphics change is judged against a
+measurement instead of an impression.
+
+It is a developer tool: available in every build, but it reconfigures the scene while it runs
+and writes files next to the client, so it is not something a player would leave on.
+
+For the per-frame overlays this builds on, see the `$details` and `$glstats` commands.
+
+---
+
+## 1. Running one
+
+Stand where you want to measure, then type in chat:
+
+```
+$bench label glp09-ring-allocator
+$bench run
+```
+
+The client freezes nothing and takes over nothing — it plays on. A line at the top of the
+screen shows which segment is running and how far along it is. When the run finishes, the chat
+prints the directory it was written to.
+
+| Command | What it does |
+|---|---|
+| `$bench list` | Prints the segment catalog and the tags you can select by. |
+| `$bench run` | Every segment, 3 repeats of 300 frames each. |
+| `$bench run fx.*` | Only the segments whose name matches the pattern. |
+| `$bench run #particles` | Only the segments carrying that tag. |
+| `$bench run fx.* 5 600` | 5 repeats of 600 frames each. |
+| `$bench quick fx.particles.off` | One repeat of 120 frames — for iterating, not for reporting. |
+| `$bench label <text>` | Names the next run. It ends up in the directory name and the report. |
+| `$bench stop` | Ends the run early. What was measured is still written out, marked aborted. |
+
+A full run is a couple of minutes. `$bench quick` is seconds and is deliberately too short to
+trust a small difference — use it to see whether a change did anything at all, then re-measure
+properly.
+
+### Getting numbers worth comparing
+
+- **Turn vsync off** (`$vsync off`). With it on, every frame time is the display refresh and
+  the client's actual cost is invisible. The report flags this, but flagging it afterwards
+  does not get the run back.
+- **Stand in the same place**, facing the same way, for both runs. The segments hold the
+  effect configuration steady; they cannot hold the world steady.
+- **Don't alt-tab.** A frame longer than half a second discards its whole repeat — the client
+  was not running normally, so what it measured was the interruption.
+- **Compare like with like.** Two runs are comparable when their manifest hash, GPU, resolution
+  and build configuration match. All four are in the report header.
+
+## 2. What a segment is
+
+A segment is one named scenario, measured on its own. Splitting a run into segments is what
+lets the result say *where* a change had an effect, rather than only whether the frame got
+faster: a change that helps particle-heavy scenes and does nothing elsewhere reads very
+differently from one that shifts every segment by the same amount.
+
+The current catalog varies the effect surfaces of whatever scene you are standing in. Each
+segment switches one surface off, so the difference between it and the baseline is what that
+surface costs at that spot.
+
+| Segment | What it isolates |
+|---|---|
+| `scene.full` | Nothing disabled. The reference every other row is read against. |
+| `fx.all.off` | All four effect surfaces at once — the upper bound. |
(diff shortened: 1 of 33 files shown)


sven-n:
[report.md](https://github.com/user-attachments/files/31576458/report.md)

It does not seem to work correctly. It runs through, but it doesn't deactivate any effects and the results of each test looks similar.

sven-n:
Thanks — the report had everything needed. It turned out to be two separate things, and one of them is a real bug.

**`fx.all.off` was broken.** `$effects off` is two actions:

```cpp
SetDisableEffects(true);
if (g_pOption) g_pOption->SetRenderAllEffects(false);
```

My segment only did the first. `RenderSprites()` and `RenderParticles()` early-return on the *options* switch and never look at `SetDisableEffects`, and `RenderJoints()` looks at neither — so the one segment that promises "everything off" left the three heaviest paths running. Your own numbers show it:

| Pass | scene.full | fx.all.off | fx.particles.off |
|---|---:|---:|---:|
| Particles draws | 170.6 | **162.4** | 0.0 |
| Sprites draws | 33.8 | **34.0** | 33.8 |
| Joints draws | 3.2 | **4.1** | 2.5 |

`SceneConfig` now carries the options switch too, and `fx.all.off` closes all three gates.

**The per-surface segments were working the whole time.** `fx.particles.off` took the Particles pass from 170.6 draws / 0.61 ms to 0.00 across 900 frames, and `fx.joints.off` is the best row in your run — 99.4 FPS against the baseline's 92.5. So the mechanism is sound; it was that one segment that lied.

**What made every row look alike is vsync.** The report has `Vsync requested / effective: yes / yes`, and every median sits between 9.89 and 10.63 ms — pinned at the refresh interval. Under that ceiling a genuine 0.6 ms saving reads as a rounding error. The report warned about it afterwards, which is precisely when the warning is useless, so it is now said in chat as the run starts, while there is still time to stop and `$vsync off`. Worth re-running that way — I'd expect `fx.particles.off` and `fx.joints.off` to separate clearly.

**The deeper problem** is that a broken segment produced a plausible-looking row instead of an obvious failure. New `segment-inert` rule: it compares each segment's draw calls per frame against the baseline and says so when nothing measurably changed. It would have flagged `fx.all.off` in your run, and it also catches the honest version of the same trap — `fx.wingshadow.off` with no wings equipped, whose 95.1 FPS is otherwise easy to read as "wing shadows are cheap" when the segment simply had nothing to disable.

Two tests cover the defect directly: every non-baseline segment must disable something, and `fx.all.off` must close all three gates.

Pushed as 22bb352.

---

PR #577, opened 2026-08-28
Add macOS (arm64) support

nyyhao:
The `APPLE` branch in `src/CMakeLists.txt` already points at
`source/App/Platform/macOS/main.mm`, but that file isn't in the tree, so
configuring for macOS stops on a missing source. I wanted to see how far the
client actually was from running there, so I added it and worked through what
came next. It turned out to be closer than I expected — the RHI/SDL3 work has
done most of the heavy lifting, and everything here is small platform plumbing
rather than anything structural.

Builds and runs on an M4 (macOS 26.5.1, arm64). It reaches the server list from
an OpenMU server and gets into the world.

### Entry point and build

- Added `App/Platform/macOS/main.mm`, mirroring the Linux entry — off Windows
  `WinMain` is an ordinary function, so the platform entry only forwards into
  it. It's Objective-C++ so Cocoa work can land there later without touching
  the build, though nothing needs it yet since SDL sets up `NSApplication`.
- The `APPLE` branch listed only `main.mm` and not `Winmain.cpp`, so `WinMain`
  was undefined at link. It now pairs them the way the `UNIX` branch does.
- The .NET client library RID came from pointer width alone, which resolves to
  `win-x64` on macOS. Native AOT rejects that outright with *"Cross-OS native
  compilation is not supported"*. Picks an `osx-*` RID from the architecture
  being built, and passes `-p:ci=true` for the same PreBuild XSLT reason Linux
  already does.
- Native AOT publishes a `.dylib` on macOS, but the copy step looked for a
  `.dll` — so the publish succeeded and the build then failed copying a file
  that never existed.
- `ld64` spells the link map flag `-map`; GNU `ld`'s `-Map` is rejected with
  *"ld: unknown options: -Map"*.
- GL, GLU and turbojpeg were linked under a Linux-only guard. Added an Apple
  branch using `OpenGL.framework`, with an explicit hint for libjpeg-turbo
  since it's keg-only under Homebrew and off the default search path.

### Platform headers

- macOS has no `<malloc.h>`; the allocator lives in `<malloc/malloc.h>`.
- GL headers are under `OpenGL/` rather than `GL/`.
- `Dotnet/Connection.h` loaded the library by a hardcoded `.so` name and
  resolved its directory through `/proc/self/exe`, neither of which exists on
  macOS. Uses the platform library name and `_NSGetExecutablePath`.

  Worth a note: `_NSGetExecutablePath` is declared directly rather than by
  including `<mach-o/dyld.h>`. That header defines
  `enum DYLD_BOOL { FALSE, TRUE }`, which redefines the Win32 `FALSE`/`TRUE`
  this codebase relies on in every translation unit that reaches it — I hit it
  as `BOOL* pbEqual = FALSE` failing to compile in `UI/Legacy/UIWindows.cpp`,
  a file with no obvious connection to the change.

### Data model (LP64)

Windows keeps `long` at 32 bits, so casting to `long` to fill a `LONG` field is
correct there. On LP64 `long` is 64-bit while `LONG` stays 32-bit, making those
brace initialisers narrowing conversions. GCC only warns, which is presumably
why the Linux build hasn't tripped over it; clang treats it as an error.

Fixed at the two definitions responsible rather than at ~30 call sites:
`RGBA()` now returns `DWORD` (a packed colour is 32 bits by definition) and
`CInput::GetScreenWidth/Height()` return `LONG` (they feed `RECT` and `POINT`).
The handful left are local casts. **This likely also affects a clang-based
Linux build**, so it may be worth having regardless of the macOS side.

### Testing

Built and ran `Debug` and `Release` on macOS 26.5.1 / arm64. The core context
loop lands on GL 3.3 (macOS caps at 4.1, so the 4.5 and 4.3 rungs fail and it
descends as designed). Logged in against OpenMU and played.

I have no Windows or Linux toolchain here, so I have **not** verified those
builds. The changes are inside `if (APPLE)` / `__APPLE__` guards except the two
return-type changes above, which are the ones worth a second look.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -211,7 +211,12 @@ elseif (CMAKE_SYSTEM_NAME STREQUAL "iOS")
     "Store restricted. When that work starts, add source/App/Platform/iOS/ "
     "and replace this branch.")
 elseif (APPLE)
-  set(MU_APP_ENTRY "source/App/Platform/macOS/main.mm")
+  # macOS entry (main.mm) plus the shared bootstrap/game-loop from Winmain.cpp,
+  # which compiles off Windows (issue #462) - the same pairing the UNIX branch
+  # below uses. Listing main.mm alone leaves WinMain undefined at link time.
+  set(MU_APP_ENTRY
+    "source/App/Platform/macOS/main.mm"
+    "source/App/Platform/Windows/Winmain.cpp")
 elseif (ANDROID)
   message(FATAL_ERROR "Android is not a supported build target yet: it needs "
     "an OpenGL ES renderer port, and its entry point is android_main via "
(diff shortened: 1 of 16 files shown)


nitoygo:
Thanks for the work on the macOS migration! Since the project has also moved to SDL3, it looks like some of the macOS-specific changes may already have been addressed as part of that migration. Could you take a look and see whether your changes are still applicable, or if they’ve already been resolved by the [SDL3 work](https://github.com/sven-n/MuMain/pull/572)?

PR #578, opened 2026-08-28
Avoid a GPU sync per vertex append in AppendBuffer

nyyhao:
While profiling the client on macOS I found the main thread spending about 53%
of its samples inside a single `glBufferSubData` call, blocked on a semaphore:

```
MainLoop -> RenderScene -> MainScene -> RenderMainScene -> RenderParticles
  -> EnableAlphaBlend -> IR::Flush -> RHI_GL_Impl::AppendBuffer
    -> glBufferSubData -> gldBufferSubData
      -> GLDContextRec::flushResource -> GLDContextRec::flushContext
        -> _dispatch_semaphore_wait_slow -> semaphore_wait_trap
```

`AppendBuffer`'s ring bookkeeping already guarantees the target range isn't one
the GPU can still be reading — `writeOffset` only moves forward, and the wrap
and grow branches orphan the buffer before restarting at zero. The comments
around DXP-26 and GLP-25 make that invariant explicit.

The problem is that `glBufferSubData` has no way to *express* it, so a driver is
free to synchronise defensively, and Apple's does: every append flushes the
context and waits for the GPU to drain. Because the cost scales with the number
of alpha-blended draws, it got worse as scene complexity rose — around 2 fps in
town degrading toward 0.5 as more effects came into view.

Writing the append through `glMapBufferRange` with `GL_MAP_UNSYNCHRONIZED_BIT`
states the invariant explicitly and the driver skips the wait. Same profile
after the change: `flushResource` and `flushContext` at zero samples,
`glBufferSubData` at one, and the main thread dominated by
`Cocoa_GL_SwapWindow` waiting on vsync — which is what a frame-limited renderer
should look like. Subjectively it went from a slideshow to smooth.

`glMapBufferRange` and `glUnmapBuffer` are core since GL 3.0 and so always
available at the GL >= 3.3 floor from GLP-08. Their declarations move above
`AppendBuffer` so it can reach them, which is why the UBO ring group no longer
declares them itself. `glBufferSubData` stays as a fallback if the map fails or
reports a discarded store.

**This isn't macOS-specific.** Desktop NVIDIA and AMD drivers hide the same
hazard with implicit buffer renaming, which is probably why it hasn't shown up
before, but nothing about the change is Apple-gated and any driver that
synchronises here should benefit. I'd expect it to be worth measuring on the
Intel HD 530 mentioned in the GLP-25 comments, since that's exactly the kind of
driver that tends not to rename.

### Testing

Profiled and ran on macOS 26.5.1 / arm64 (GL 3.3 core context), before and
after, using `sample(1)`. I don't have a Windows or Linux toolchain here, so I
have **not** verified those builds or measured the change there — worth a look
before merging, since this is a shared code path rather than a platform guard.

### Not addressed

The same pattern remains in text rendering:
`CUIRenderTextOriginal::UploadText` -> `RHI_GL_Impl::UpdateTexture` ->
`glTexSubImage2D` -> `prepareResourceForCPUAccess` -> `flushContext`, about 12%
of main-thread samples after this fix. Left alone here to keep the change
focused, but it's the same class of problem if someone wants it.

### DIFF
--- a/src/source/Render/RHI/RHI_GL.cpp
+++ b/src/source/Render/RHI/RHI_GL.cpp
@@ -225,6 +225,11 @@ typedef void (APIENTRY* PFNGLBUFFERDATAPROC)(GLenum target, GLsizeiptr size, con
 typedef void (APIENTRY* PFNGLBUFFERSUBDATAPROC)(GLenum target, GLintptr offset, GLsizeiptr size, const void* data);
 typedef void (APIENTRY* PFNGLDELETEBUFFERSPROC)(GLsizei n, const GLuint* buffers);
 typedef void (APIENTRY* PFNGLBINDBUFFERBASEPROC)(GLenum target, GLuint index, GLuint buffer);
+// Declared here rather than with the UBO-ring group further down because
+// AppendBuffer (above that group) streams through them on the unsynchronised
+// path; the UBO ring's loader assigns the same pointers.
+typedef void*     (APIENTRY* PFNGLMAPBUFFERRANGEPROC)(GLenum target, GLintptr offset, GLsizeiptr length, GLbitfield access);
+typedef GLboolean (APIENTRY* PFNGLUNMAPBUFFERPROC)(GLenum target);
 
 namespace {
     PFNGLGENBUFFERSPROC     fn_glGenBuffers     = nullptr;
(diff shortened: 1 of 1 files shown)


nitoygo:
Tested in windows. No noticeable or visible perf difference in Windows.

PR #579, opened 2026-08-28
Only shake the camera for a nearby source

nyyhao:
### What I saw

Playing in **Lorencia**, walking around town, the camera would start juddering
hard — like a constant earthquake — for a few seconds, stop, then start again a
little later somewhere else in the city. It was bad enough to be unplayable in
places, and it didn't happen out on the hunting maps.

Setup was the MuMain client on macOS against a local **OpenMU** server with its
bot population enabled (50 bot characters), which is what made it so obvious:
bots warp into Lorencia to shop and fight, so there was almost always somebody
casting something nearby.

### Tracking it down

I logged `EarthQuake` where `DefaultCamera` applies it, along with the hero's
tile. It fired around tiles (139–144, 131–137) with values between `-0.6` and
`-0.1` in `0.1` steps. That range identifies the source exactly — it's
`(rand() % 6 - 6) * 0.1f` in `Move_MODEL_CIRCLE_LIGHT`, and across 88 samples I
never saw `-0.7` or `-0.8`, which rules out the similar-looking
`(rand() % 8 - 8) * 0.1f` sites in `ZzzEffect.cpp`.

Working back from there: `MODEL_CIRCLE_LIGHT` at `SubType` 0 comes from
`AT_SKILL_HELL_FIRE` in `ZzzCharacter.cpp`. So it was other players' Hellfire
casts. The handler sets `EarthQuake` on every frame the effect is alive, and
`EarthQuake` is a single global that `DefaultCamera` adds straight onto the
camera pitch — with no distance test anywhere in that path. A caster shakes
every nearby player's view at full strength no matter how far away they are,
including sources off-screen entirely.

Decay is `EarthQuake *= 0.2f` once per frame in `InitializeSceneFrame`, so it
dies quickly on its own; the problem is purely that something in scope keeps
re-setting it.

### The change

`SetEarthQuakeAt()`, declared next to the global it writes, applies the shake
only when its source is within about twelve tiles — roughly a screen — and
scales it linearly to zero at that edge. Magnitude at zero distance is
unchanged, so an effect feels exactly the same to whoever is standing in it;
what stops is strangers across the map shaking your camera.

Routed the seventeen object-driven sites in `ZzzEffect`, `EffectBehaviors`,
`MoveHandlers` and `GOBoid` through it. The scripted map-wide shakes in
Raklion, Hellas, Kanturu, Chaos Castle and Cursed Temple are deliberately left
alone — those are meant to shake the whole map during an event, and several
have no local object to measure a distance from.

### On whether you want this

This is a **behaviour change**, not a port fix, so it's entirely your call. The
argument for it is that a screen shake from a caster you can't see is hard to
defend on any reading. The argument against is that it's long-standing
behaviour and someone may consider the current feel correct — it's only really
punishing when a map is busy, which a bot-populated test server exaggerates far
beyond a normal server.

Happy to drop it, narrow it to fewer effects, or make the radius configurable
if you'd rather.

### Also noticed, deliberately not changed

`(rand() % 6 - 6) * 0.1f` yields `-0.6 … -0.1` — always negative, never zero or
positive. So it isn't a shake around a centre, it's a persistent downward pitch
with jitter on top, and every one of my 88 samples was negative. Several other
sites share the shape (`% 8 - 8`, `% 4 - 4`). It looks like an off-by-one, but
fixing it would change how every one of those effects looks, so I left it well
alone.

### Testing

macOS 26.5.1 / arm64, GL 3.3 core, against OpenMU with 50 bots. No Windows or
Linux toolchain here, so those builds are unverified — this touches shared
files rather than sitting behind a platform guard.

### DIFF
--- a/src/source/Engine/AI/GOBoid.cpp
+++ b/src/source/Engine/AI/GOBoid.cpp
@@ -759,7 +759,7 @@ void RenderDarkHorseSkill(OBJECT* o, BMD* b)
                 CreateEffect(MODEL_GROUND_STONE + rand() % 2, Position, o->Angle, o->Light);
             }
         }
-        EarthQuake = (rand() % 3 - 3) * 0.7f;
+        SetEarthQuakeAt(o->Position, (rand() % 3 - 3) * 0.7f);
     }
     else if (o->WeaponLevel == (BYTE)(19.f / FPS_ANIMATION_FACTOR))
     {
@@ -802,7 +802,7 @@ void RenderSkillEarthQuake(CHARACTER* c, OBJECT* o, BMD* b, int iMaxSkill)
 
     if (o->WeaponLevel == iMaxSkill - 1)
     {
-        EarthQuake = (rand() % 3 - 3) * 0.7f;
+        SetEarthQuakeAt(o->Position, (rand() % 3 - 3) * 0.7f);
         CreateEffect(MODEL_SKILL_FURY_STRIKE, TargetO.Position, TargetO.Angle, TargetO.Light, 0, o, -1, 0, 2);
     }
 
--- a/src/source/Engine/Object/ZzzObject.cpp
+++ b/src/source/Engine/Object/ZzzObject.cpp
@@ -69,6 +69,22 @@ OPERATE      Operates[MAX_OPERATES];
 //int   World = -1;
 float EarthQuake;
 
+void SetEarthQuakeAt(const vec3_t sourcePosition, float magnitude)
+{
+    if (Hero == nullptr) return;
+
+    // ~12 tiles: roughly what fits on screen, so anything that can shake the view
+    // is something the player can actually see happening.
+    const float kShakeRadius = TERRAIN_SCALE * 12.f;
+
+    const float dx = sourcePosition[0] - Hero->Object.Position[0];
+    const float dy = sourcePosition[1] - Hero->Object.Position[1];
+    const float distSq = dx * dx + dy * dy;
+    if (distSq >= kShakeRadius * kShakeRadius) return;
+
+    EarthQuake = magnitude * (1.f - sqrtf(distSq) / kShakeRadius);
+}
+
 static  int g_iThunderTime = 0;
 int g_iActionObjectType = -1;
 int g_iActionWorld = -1;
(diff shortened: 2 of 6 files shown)


Mosch0512 (commented):
Thanks for tracking this down. Tying the sample range to `Move_MODEL_CIRCLE_LIGHT` and then to Hellfire made this easy to follow. I'm in favour of the change: a camera shake from a caster you can't see is hard to justify. `SetEarthQuakeAt` is a clean way to fix it, and keeping full strength at distance zero means the caster's own feel doesn't change.

There's one gap that I think should be fixed before merge, plus one small thing:

**1. Far effects can still cancel a nearby shake** (`MoveHandlers.cpp` 7550, 7651, 8980)

`Move_MODEL_BLOW_OF_DESTRUCTION` (SubType 0 and 2) and `Move_MODEL_EMPIREGUARDIAN_BLOW_OF_DESTRUCTION` switch between two branches:

```cpp
if (o->LifeTime >= 15.f)
    SetEarthQuakeAt(o->Position, ...);
else
    EarthQuake = 0.f;   // still global, no distance check
```

The "on" branch is now range-limited, but the "off" branch still resets the global for everyone. Say you're standing in a Hellfire and a Blow of Destruction finishes somewhere else on the map. If that effect updates after the Hellfire in the same frame, your shake is cancelled. A far source can no longer start a shake, but it can still stop one, which is the same kind of cross-talk this PR is trying to remove. A matching helper that only clears when the source is in range would fix it, or those branches could just skip the reset when out of range.

**2. Out-of-range sources still damp the shake** (`MoveHandlers.cpp` 6138, 6345, 6502)

`EarthQuake *= pow(1.0f / 2.f, FPS_ANIMATION_FACTOR);` runs right after `SetEarthQuakeAt`, even when the source was too far away and nothing was set. That halves a shake some other nearby source caused. It's minor, but it would be tidy to only damp when the call actually set a value. For example, `SetEarthQuakeAt` could return a `bool`.

Apart from that it looks good to me. I agree with leaving the scripted map-wide shakes alone, and with not touching the always-negative `rand() % N - N` ranges in this PR. That's worth its own follow-up.

PR #591, opened 2026-09-15
bugfix: send CloseNpc on Kanturu gate dialog close

eduardosmaniotto:
## Summary

Fixes stuck NpcDialogOpened after using the Kanturu 2nd→3rd gate dialog (0xD1), which never sent CloseNpcRequest. Adds ClosingProcess() and hooks it in Hide(), same as 0e89b67.

## Related issues

NA

## Checklist

- [X] I have built the project locally (see [`docs/build/README.md`](docs/build/README.md)).


Mosch0512 (changes requested):
The fix's approach is sound for the paths it covers, but two verified gaps mean the underlying "stuck NpcDialogOpened" bug can still be reproduced:

- Dying during the Enter animation (frame 0–42) skips both the enter request and any close call, leaving the server dialog stuck open — confirmed by tracing GM_Kanturu_2nd.cpp:1232-1249 and the network handler chain.
- HideAllGroupA()/HideAllGroupB() hide the dialog directly, bypassing the class entirely — most triggers (inventory/character/party keys) turned out to be gated off, but talking to another NPC (ReceiveNPCDlgUIStart) isn't, and still reproduces the bug.
- A minor convention note: the diff touches the already ~96-line BtnProcess() without extracting the four near-identical failure-branch blocks, per the repo's CODING_RULES.md Rule 1.

eduardosmaniotto:
Thanks for the detailed trace — all three points addressed:
1. Death during Enter animation (frames 0–42) — fixed in GM_Kanturu_2nd.cpp.
You're right: the Hero->Dead guard at frame 42 skips the enter, and the frame-50 reset then cleared the animation with no Close ever sent. I added a guard in the >= 50 block: if IsEnterRequest() == false (enter never went out), it calls ClosingProcess() before resetting. When the enter was sent, the response path (ReceiveKanturu3rdEnter) still owns the close, so no double-Close. I kept it in the animation block rather than hooking ReceiveDie — a networking→UI call there would cross layers, and death in Kanturu keeps the gate object alive so frame 50 is reached.
2. HideAllGroupA/B bypass — fixed, but at the click site rather than in HideAll.
I initially put the Close in HideAllGroupA/B and then reverted it: a Close fired from Show(NPC_DIALOGUE) necessarily lands on the server after the already-sent Talk(B), so it would clear B's just-opened dialog instead of the stale gate one. The fix is in ZzzInterface.cpp (Kanturu2nd branch): when the gate dialog is visible and no animation is running, ClosingProcess() is sent before SendTalkToNpcRequest, keeping Close(old) → Talk(new) order. This holds under both replace and reject-while-open server semantics — and in the reject case it actually unsticks the server too. Also added a NOTE at the Hide() chain documenting why Kanturu deliberately has no hook there (it runs on the Enter path, where a Close would drop the pending enter at frame 42).
3. Rule 1 — done.
Extracted RejectEnterRequest(BYTE) (CreateMessageBox + ClosingProcess); the four failure branches now one-line it.

Mosch0512 (approved):

PR #593, opened 2026-09-17
fix(editor): do not build an editor singleton while shutting down

fedetandil:
## The problem

`CMuEditorCore::Shutdown` saves the item and skill editors' column preferences unconditionally:

```cpp
g_MuItemEditorUI.SaveColumnPreferences();
g_MuSkillEditorUI.SaveColumnPreferences();
```

Both accessors are Meyers singletons that build their instance on first use, so asking for one that was never opened **constructs it right there**.

`Shutdown` is reached from `~CMuEditorCore`, which runs during static destruction, and those constructors read `g_MuEditorConfig` — another static in a different translation unit, whose destruction order relative to this one is unspecified. When it has already been destroyed, the read lands in freed memory and the process segfaults on the way out.

It only reproduces when the editor was never opened during the session, which is the common case, and it happens after everything the user cares about has already finished — so it reads as a crash on exit with no obvious cause.

## The fix

Add `HasInstance()` to both editors, tracked by a flag set inside `GetInstance()`, and save only what was actually constructed. An editor that was never opened has no preferences worth writing, so nothing is lost.

## Notes

- Built and verified on Linux (SDL3 / SDL_GPU).
- No behaviour change when the editor *was* opened: both paths still save exactly as before.


src/MuEditor/Core/MuEditorCore.cpp
```diff
@@ -362,11 +362,17 @@ void CMuEditorCore::Shutdown()
     if (!m_bInitialized)
         return;
 
-    // Save item editor preferences before shutting down
-    g_MuItemEditorUI.SaveColumnPreferences();
+    // Constructing an editor here would run during static destruction and read an already
+    // destroyed g_MuEditorConfig, so only save the ones that were opened.
+    if (CMuItemEditorUI::HasInstance())
```
Mosch0512: This still has the use-after-destroy problem when the editor **was** opened.

Function-local statics are destroyed in reverse order of construction. `g_MuEditorCore` is constructed first, `g_MuEditorConfig` later (in `Initialize`, via `g_MuEditorConfig.Load()`), and the item/skill editor singletons later still (when the user opens them). So by the time `~CMuEditorCore` → `Shutdown()` runs during static destruction, both the editor and the config have already been destroyed. `HasInstance()` is still `true` though, so `SaveColumnPreferences()` reads the destroyed `m_columnVisibility` and writes into the destroyed config.

This only happens on the static-destruction path (on Windows, `Winmain.cpp` calls `g_MuEditorCore.Shutdown()` explicitly first).

Suggested fix: clear the flag in `~CMuItemEditorUI` / `~CMuSkillEditorUI`, or just drop these two calls. Each editor's destructor already calls `SaveColumnPreferences()`, so the calls here add nothing.

PR #594, opened 2026-09-17
Developer control socket: drive a client from a shell, behind ENABLE_CONTROL_SOCKET

alandarev:
## Summary

Client-side work is hard to test: every check needs a human at the keyboard,
and anything that takes two clients needs two. This adds a **developer control
socket** — a local stream socket (`AF_UNIX`, the same on Windows and on Linux)
that a script can drive the client through: log in, pick a character, walk,
fight, cast, pick up, chat, press a key, click a pixel, take a screenshot, and
read back what the client was told by the server.

It is **off at build time**. `ENABLE_CONTROL_SOCKET` defaults to `OFF` and
mirrors `ENABLE_EDITOR`: a player build compiles none of the feature, and the
`MU_CONTROL_SOCKET` variable name does not exist in the binary, so there is no
scripting API to switch on. The plain presets (`linux-x64`, `windows-x86`,
`windows-x64`, `macos-arm64`) and CI stay player builds; the `-mueditor`
presets turn it on, and a `linux-x64-mueditor` pair is added to mirror the
Windows ones.

## What is in the diff

| Folder | What |
|---|---|
| `src/source/GameLogic/Automation/` | the MU Helper's mouse-free primitives (`SimulateMove`, `SimulateBasicAttack`, `SimulateSkill`, `ObtainItem`) extracted as free functions with result enums, so the helper keeps its policy and other callers can reuse the rules. Behaviour-preserving; `CMuHelper` now calls them |
| `src/source/Core/Platform/LocalSocket.*` | non-blocking local stream socket for Windows and Linux alike: stale-file replacement, owner-only (`0600`) creation, unlink on close, line buffering |
| `src/source/Core/Text/Utf8.*` | the small code-page conversion the protocol needs (and the `<windows.h>` include it takes to build on Windows) |
| `src/source/Core/Input/SyntheticInput.*` | one injected key or click at a time, below the game's own readers |
| `src/source/App/Control/` | newline-delimited JSON protocol, dispatch table, the session/world/query command sets, the 2,048-entry event ring and the recorders |
| `src/source/{Scenes,UI/Windows}/` | entry points extracted from the click paths (`CServerSelWin::SelectServer`, `CLoginWin::SubmitCredentials`, `CMsgWin::PendingMessageCode`/`DismissMessage`, `Scenes::StartGameWithSlot`) so both routes share one implementation |
[...]

PR #580, opened 2026-09-01
Rmlui on sdl gpu

nitoygo:
Depends on https://github.com/sven-n/MuMain/pull/572 

## Summary

Modernize UI implementation of MuMain using RmlUI.  
Allows for custom themes, while preserving the legacy theme choosable.

## Related issues

N/A

## Checklist

- [ ] I have read and followed [`docs/CODING_RULES.md`](docs/CODING_RULES.md).
- [ ] I have built the project locally (see [`docs/build/README.md`](docs/build/README.md)).
- [ ] Changes are focused on one concern; the diff is as small as it reasonably can be.
       Note: The changes are focused on one concern, but the diff is large as it overhauled the UI

PR #596, opened 2026-09-18
dev/rmlui-ui-system + developer control socket (#594) merged, for testing

alandarev:
Draft, for testing only — the successor of nitoygo/MuMain#4 on the new branch. This is #595 plus a merge of `feature/control-socket` (#594, the developer control socket behind `ENABLE_CONTROL_SOCKET`) into `dev/rmlui-ui-system`, so that scripted clients can drive the RmlUi tree: login/select-char over the socket, `screenshot`, `hotkey`, `click-ui`, and the automation we use to compare the ported windows against the original UI (same scenarios on both builds, paired screenshots, pixel diffs).

Not meant to be merged as-is: #594 is the upstream PR for the socket, and this merge exists so both can be tested together on this branch. The merge commit message lists every conflict resolution (eight files: `src/CMakeLists.txt`, `Winmain.cpp`, `WSclient.cpp`, `LoginWin`, `ServerSelWin`, `MsgWin`, plus the socket's own files adapted to the post-#580 header/namespace layout).

Verified: `cmake --preset linux-x64 -DENABLE_CONTROL_SOCKET=ON`, release build, `ctest` **237/237**; a scripted client logs in, enters the world, swaps themes and takes screenshots on this tree.


alandarev:
Pushed two things:

1. the merge of upstream `18c20b7f` (so this branch is current again);
2. **a real fix, not just a merge**: `fix(control): inject clicks and keys as SDL events, so RmlUi sees them too`.

The second one is the adaptation this merge turns out to need. On `main` the socket's `click-ui`/`hotkey` are complete — they write the legacy mouse globals and the key-state shim, which is the whole of that client's input. On this branch they are half of it: RmlUi documents are driven exclusively by SDL events (`RmlUiRuntime::ProcessSdlEvent`), so an injection written straight into the legacy globals does not exist as far as RmlUi is concerned. Concretely, before this commit a scripted client could not press the options window's tabs, its dropdowns, or its close button — I hit it while testing the UI-scale row in #598.

The injector now pushes real motion/button and key-down/key-up events onto SDL's own queue, so the event loop does what it does for physical input: the UI consumer gets first refusal, the legacy readers get what it leaves unclaimed. Both tiers work, and the arbitration between them is real rather than bypassed — the old path fired a legacy click even with the pointer over an RmlUi panel.

Kept deliberately: the press/held/release frame staging (the legacy readers consume edges per rendered frame, so a pair delivered in one pump can be swallowed) and the `IsKeyHeld` shim (SDL's own keyboard/mouse state reports physical devices and is not moved by a pushed event). A click is now preceded by a pointer move to its own position, since a press is dispatched to whatever the pointer last moved over.

Verified live on this branch: `click-ui` opens the RmlUi system menu from the HUD button, walks it to the options window, switches its tab, opens the UI Theme dropdown and commits an option; `hotkey esc` closes the window, `hotkey i` opens the inventory. `ctest` 238/238, clang-format and cppcheck clean.

If you would rather have this on `main` instead (as part of #594), say so — it works there too, it is just not needed there, and the failure it fixes is only visible on this branch.

alandarev:
Follow-up pushed: **`feat(control): type, for the text fields a key press cannot fill`**.

The previous commit made injected keys real SDL key events, which is what RmlUi needs — but a key press carries no character in this client. Characters arrive as their own text event, which is what a portable text field and an RmlUi input actually read, so `hotkey` could move the caret in a focused field and never put anything into it. Every flow that needs a typed name — account and password, a character name, a guild name — was out of reach of a scripted session.

`type` (`text`, optional `enter`) sends the characters to whatever has keyboard focus and, with `enter`, follows them with a Return press one frame later for fields that submit on it. Printable text only, at most 256 bytes; a newline or a tab is a `hotkey`, not field content. UTF-8 passes through unchanged. The injection owns the string until a frame after the event is consumed, since SDL copies a pushed event but not the string it points at.

Verified live on the RmlUi login dialog, without touching the machine's keyboard: click the account field, `type test1` → the characters appear; `hotkey tab`, `type --enter <password>` → the password lands in the second field and the Return submits it, the server answers. `ctest` 240/240, clang-format and cppcheck clean.

Docs updated in the same commit: the "Not covered: typing text" line is gone, replaced by what `type` does.

alandarev:
Caught up with the four fixes from the #594 review (optional act deadline so `events --follow` survives, the guarded server-group button index, slot-only `pickup`, self-named scripted screenshots). The only branch-specific difference is the `SelectServer` guard: this window has no button array, so the position is bounded against `kMaxGroups` and only the selected index is stored. Rebuilt with the socket: `ctest` 240/240, `tools/check client` clean.

nitoygo:
Merged latest main to dev/rmlui-ui-system instead

PR #597, opened 2026-09-19
dev/rmlui-ui-system: linear ViewportFitScale so UIScalePercent=100 reproduces the legacy layout

alandarev:
Slice 1 of the UI-scale change we discussed: make the auto-fit ramp linear again, so `UIScalePercent=100` lands the migrated UI on exactly the pixels the original client draws at every resolution. Slice 2 (the `UI scale` dropdown on the options window's UI tab, 50–200 %, live apply, tooltip) follows on top of this.

## Why

`ViewportFitScale` had a quadratic damping (`kFitDampingExponent = 2.0`, 9798a154) between the 640×480 reference and each ceiling. Only the endpoints matched the original; every ordinary resolution in between came out smaller — 1.25× instead of 1.5× at 1280×720, 1.36× instead of 1.6× at 1024×768. That makes the `legacy` theme's pixel-parity claim untestable at any real resolution and lets the still-native sprites (never damped) drift from the RmlUi documents beside them.

The reason the damping was added — "too upscaled at my resolution" — is what the `UIScalePercent` dial is for: it multiplies the fit post-clamp, so the curve every window shares stays the one predictable formula and the user turns the dial down instead.

## What

- `UITransform.cpp`: `ViewportFitScale` = `clamp(min(w/640, h/480), 1, ceiling)`; constant and damping block removed. Stale mentions in `UITransform.h` and `TitleSceneUI.cpp` cleaned up.
- `tests/ui/test_ui_scaling.cpp`: the seven assertions that 9798a154 recomputed for the damped curve go back to their pre-damping values.
- `docs/rmlui-ui-system/layout-and-scaling.md`: states the ramp is linear and why.

## Verification

- Linux player build, `ctest` **210/210**; clang-format / cppcheck range check clean.
- Live, 1280×720, legacy theme, world render off, same map position, against `main` (with the bottom HUD bands centred the way this branch already does): HUD bar x-extent **160–1119 (main) vs 160–1117 (this branch)** — before this change the branch's bar was 240–1040.

Bottom HUD strip, main on top, this branch below:

HUD at 1280x720: main (top) vs dev/rmlui-ui-system with this fix (bottom)

Full frames: [pr597-full.png](https://raw.githubusercontent.com/alandarev/mumain-sven/assets/pr-597/pr597-full.png).

PR #598, opened 2026-09-19
dev/rmlui-ui-system: a UI Scale row in the options window's UI tab

alandarev:
Stacks on #597 (linear `ViewportFitScale`) — the diff here is only the second commit; the base will fast-forward once #597 merges.

`[UI] UIScalePercent` has been live since the `dp` wiring landed, but the only way to change it was hand-editing `config.ini` and restarting. This adds the in-game control it was always meant to have, next to the UI Theme row it belongs with.

UI Scale row, legacy theme, at 70 %

## What it does

- **A 6th `.option-dropdown`** (id 5) on the UI tab, over a fixed ladder — 50/60/70/80/90/100/125/150/200 %, with 100 % selectable so "back to normal" is one click. The ladder and the "which entry do I show for an arbitrary config value" lookup are free functions in `UI::Options`, so both are testable without standing up the window.
- **Applies immediately**, the way the other rows on that tab do: persist, then resize the window to the size it already has (`MuApplyWindowResolution(WindowWidth, WindowHeight, windowed)`). Nothing recomputes the ratio on its own, but every resolution-dependent system does so on a resize — all three contexts' `dp` ratio, `UI::Scaling`'s active transform, the legacy `CWin` layout, the 3D UI cameras. Unlike `ApplyResolution()` this deliberately does **not** write `[Window] Width/Height` back; the window size did not change. The apply is deferred to `Update()`, out of RmlUi's own event dispatch, for the same reason the theme switch beside it is.
- **`GameConfig::SetUIScalePercent()` now clamps to 50–300 %** (new `CfgMin/MaxUIScalePercent`). The floor is a usability limit, not a technical one: below it the options window's own rows get too small to click the setting back up. A value read from disk is still taken as-is, so a hand-edited `config.ini` between two offered steps keeps working and the row shows the nearest entry.
- **The row carries this window's first hover tooltip** ("Multiplier on top of the automatic fit to the window size") — the label alone cannot say what the percentage multiplies. It is a plain sibling span shown by an RCSS `:hover` rule, the same declarative mechanism `main_frame.rcss`'s gauge tooltips use; no C++ hover callback, which `CMainFrameWindow`'s skill tooltip only needs because its content is per-skill data. In the modern theme it reuses `base.rcss`'s shared `.tooltip` shell; the legacy theme has no such shell, so its own `option_window.rcss` supplies the look.
- Being the **last row of its tab**, this row's dropdown list opens *upwards* and its tooltip sits *below* the row: a 9-entry list opening downwards painted over the panel's bottom frame art and the close button (seen live, fixed before committing).

`RmlDropdownOptionClick()`'s switch is expanded to one statement per line because CI's clang-format check rejects a short case label on its own line and `case 5` made the whole block a changed range.

## Verification

- Player build clean; `ctest` **212/212** (210 before — two new cases cover the ladder's invariants, the nearest-entry lookup and the clamp); clang-format + cppcheck on the changed ranges clean.
- Live against a real server at 1280×720, **both themes**: picking 70 % and 150 % rescales everything immediately, no crash; switching themes afterwards still works.
- The dropdown path is **pixel-identical** to setting the same percentage by hand and restarting — a 70 % frame reached through the dropdown and a 70 % frame reached by re-applying the setting differ in zero pixels, and going back to 100 % reproduces the original 100 % frame exactly.
- `config.ini` after the session is byte-identical to before it except `UIScalePercent`; `[Window]` is untouched.

Screenshots of both themes (row, tooltip, open list, 150 %) in the comment below.


alandarev:
All shots are 1280×720, a live session in Lorencia, Options → **UI** tab, taken with the client's own screenshot command. The legacy shots are at `RmlTheme=legacy`, the modern ones at `RmlTheme=modern`; the 3D world is switched off in most of them so only the UI is drawn. The headline image in the description is the same window during ordinary play.

<details><summary>All 7 shots (click to expand)</summary>

**Legacy — the new row, at 100 %**
legacy 100

**Legacy — hover tooltip** ("Multiplier on top of the automatic fit to the window size", wrapped to two lines under the row)
legacy tooltip

**Legacy — the list open** (upwards, since this is the last row of the tab; the current entry is highlighted, the list scrolls)
legacy list

**Legacy — 70 % applied** (whole UI, HUD included, rescaled immediately)
legacy 70

**Modern — the new row, at 100 %**
modern 100

**Modern — the list open, with the tooltip below the row** (the tooltip uses `base.rcss`'s shared `.tooltip` shell here)
modern list

**Modern — 150 % applied** (at 1280×720 the panel then outgrows the window; that is what the 200 % ceiling is about)
modern 150

</details>

alandarev:
Rebased onto `dev/rmlui-ui-system` now that #597 is merged — the conflict is gone and the diff is this one commit again.

Two things came out of the rebase, both folded into the same commit:

- `OptionWindow.h`'s dropdown-id comment conflicted with the pruning in `18c20b7f`; it keeps your shortened wording, with `5=uiScale` added.
- The new comments this change introduces were written in the older journal style, so I reworded them to match `18c20b7f`: present tense, no "found live", no dates or doc pointers.

Re-verified after the rebase: player build clean, `ctest` 212/212, clang-format and cppcheck clean on the changed ranges. The screenshots in the gallery comment above are from the pre-rebase build; nothing about the UI changed, only comments.

PR #599, opened 2026-09-20
fix(render): report SDL video driver and error on init failure

r0naldoom:
## Summary

When `SDL_InitSubSystem(SDL_INIT_VIDEO)` fails, `MuError.log` only says `> SDL video init failed.` and the process exits with code 0 after the dialog. This adds the requested video driver and `SDL_GetError()` to that line. Nothing else changes.

## Root cause

SDL loads its Linux video backends (X11, Wayland, libdecor, KMS) dynamically at run time. In a packaged runtime where those libraries are not resolvable from the executable, video initialization fails before any window exists, and the current log line does not say which driver was tried or why it failed. The same line is the only evidence a user can attach when a Linux startup report is filed.

## Fix

`src/source/App/Platform/Windows/Winmain.cpp:1985` (this file is the entry point on Linux too): read `SDL_GetHint(SDL_HINT_VIDEO_DRIVER)` (or `auto` when unset), convert `SDL_GetError()` with the `Utf8ToWide` helper the file already uses, and write both through the existing `g_ErrorReport.Write`. The dialog and the `return 0` are unchanged.

## Related issues

None open upstream; groundwork for reproducible Linux/Wayland startup reports.

## Test plan

- [x] Linux x64 Release: `cmake --preset linux-x64 && cmake --build --preset linux-x64-release && ctest --test-dir out/build/linux-x64 -C Release`: clean build, 206/206 tests pass.
- [x] `clang-format --dry-run --Werror` (21.1.8) and `cppcheck` with the CI arguments on the changed file: clean.
- [x] Failure path, Linux runtime where SDL's dynamically loaded Wayland backend is not resolvable, `SDL_VIDEO_DRIVER=wayland ./Main`:
      `> SDL video init failed. Requested driver: wayland. SDL error: wayland not available`
- [x] Success path unchanged, same tree configured with `-DSDL_WAYLAND_SHARED=OFF -DSDL_X11_SHARED=OFF -DSDL_WAYLAND_LIBDECOR_SHARED=OFF -DSDL_KMSDRM_SHARED=OFF` (backends linked), `SDL_VIDEO_DRIVER=wayland ./Main`:
      `> Start window success.` / `SDL_gpu -- device driver: vulkan`
- [ ] Windows and macOS: not built here. The change is a log line on the failure path; the Windows `MessageBox` and return value are untouched.

## Checklist


sven-n:
## Code review

Reviewed against [`docs/CODING_RULES.md`](../blob/main/docs/CODING_RULES.md). **No issues found — merging.**

What I checked:

- `Utf8ToWide` is defined unconditionally at `Winmain.cpp:1141` (no `#ifdef` guard), well above the call site, so the Windows build sees it too.
- `%ls` in a `wchar_t` format string is the established convention for `g_ErrorReport.Write` in this codebase (`GlobalBitmap.cpp:663`, `:796`, `:879`, …), and it is the correct conversion for a `wchar_t*` argument in wide-character `printf` on both MSVC and glibc.
- `SDL_GetHint(SDL_HINT_VIDEO_DRIVER)` returning `nullptr` when unset is handled, and the empty-string case is handled separately — good, since `SDL_VIDEO_DRIVER=` would otherwise print an empty driver name.
- `SDL_GetError()` is read immediately after the failed `SDL_InitSubSystem`, before anything else can overwrite the thread's error slot.
- Blast radius is one log line on a path that already terminates the process. The `MessageBox` and the `return 0` are untouched, as stated.

Two things noted but deliberately not asked for here, since both are pre-existing and outside this PR's one concern:

- `CErrorReport::Write` formats into a fixed 1024-`wchar_t` buffer with raw `vswprintf` rather than the `vswprintf_s` wrapper in `Core/Platform/SecureCrt.h`. SDL error strings are short enough that this line will not hit it.
- The `return 0` on a fatal init failure still reports success to the shell, which your description calls out. Worth a separate change if you want Linux packagers to be able to detect the failure from the exit code.

---

PR #600, opened 2026-09-20
feat(log): route SDL logs to MuError.log

r0naldoom:
## Summary

Routes SDL output to `MuError.log` without removing the existing stderr output. SDL priorities map to a corresponding spdlog severity, and each stored line uses a readable SDL category name.

## Root cause

The application initializes `mu::log`, but never registers an SDL log-output callback. SDL therefore uses its default stderr sink only. Diagnostics such as the compositor fifo-v1 fallback and failed X11 connection are absent from the persistent log a user can attach to a report.

## Fix

`src/source/Core/Utilities/Log/SdlLogBridge.{h,cpp}` owns the SDL-to-spdlog bridge in `Core::Log::Sdl`. Its scoped installer captures the previously installed SDL callback and userdata with `SDL_GetLogOutputFunction`, forwards each message to that callback and to the attachable `sdl` logger, and on destruction restores the previous callback first and releases the logger reference second, so the last messages before teardown still reach the forwarder. The forwarder takes no lock: SDL holds its log-function lock across the callback and across `SDL_SetLogOutputFunction`, which already serializes invocations and guarantees none is running once the callback is restored; a comment on `ForwardSDLLog` records that constraint. A nested instance stays inert instead of disarming the outer one. The named forwarding level deliberately retains the diagnostics selected by `SDL_LOGGING`.

`Winmain.cpp` holds the installer in a `std::optional` created right after `mu::log::Init()`; every early return restores SDL's callback through the destructor, and `ShutdownRuntime()` releases it after `SDL_Quit()` and before `mu::log::Shutdown()`, matching the nesting the regression test uses.

`docs/build/linux/console.md` documents `MuError.log` and `SDL_LOGGING` for reporters.

## Related issues

None open upstream; captures existing SDL diagnostics in the application's attachable log.

## Test plan

- [x] Linux x64 Release: `cmake --preset linux-x64 && cmake --build --preset linux-x64-release && ctest --test-dir out/build/linux-x64 -C Release --output-on-failure`: clean build, 207/207 tests pass.
- [x] `clang-format --dry-run --Werror` (21.1.8) and `cppcheck` with the CI arguments on the changed source files: clean.
- [x] `SDL bridge writes named diagnostics to MuError.log`: installs the scoped bridge, emits an SDL video info message, creates and destroys a nested guard, emits a second message, explicitly shuts down `mu::log` after scope exit, and verifies both persisted `[sdl] [info] video:` records.
- [x] COSMIC automatic run with `SDL_LOGGING='video=verbose,gpu=verbose' timeout --signal=TERM 60s ./Main`:
      `MuError.log: [sdl] [info] video: This compositor lacks support for the fifo-v1 protocol; falling back to XWayland for GPU performance reasons`
      `MuError.log: [sdl] [debug] video: Failed to connect to the X11 display server`
      Both lines also remain on stderr; startup continues to `> Start window success.`
- [x] Dynamic-backend failure path: `SDL_VIDEO_DRIVER=wayland timeout --signal=TERM 60s ./Main` reaches the existing video-initialization failure. SDL produces `wayland not available` with `SDL_SetError`, which does not invoke an SDL log-output callback; this focused sink deliberately leaves that caller-side error reporting to the separate startup diagnostic change.
- [ ] Windows and macOS: not built here. The installer uses SDL's cross-platform callback API and is covered by the release test.
- [ ] The three early-return paths with a live SDL logging thread are not exercised; the guarantee there rests on SDL's lock rather than on anything this PR adds.

## Checklist


sven-n:
## Code review

Reviewed against [`docs/CODING_RULES.md`](../blob/main/docs/CODING_RULES.md). The mapping itself is correct and the idea is right — SDL's diagnostics belong in the attachable log. Four things to address before this goes in.

### 1. Install and restore are not paired (correctness)

`InstallSDLLogOutput()` runs inside `InitializeWorkingDirectoryAndLog()`, but `RestoreSDLLogOutput()` is only reached from `ShutdownRuntime()` and from the one font-failure path at `Winmain.cpp:2187`. `WinMain` has three other exits after logging is up that neither restore the callback nor call `mu::log::Shutdown()`:

- `Winmain.cpp:1987` — SDL video init failed, `return 0` (the line PR #599 touches)
- `Winmain.cpp:2002` — `SDL_CreateWindow` failed, `return 0`
- `Winmain.cpp:2035` — `mu::InitSDLGpuRenderer` failed, `return FALSE`

On all three, SDL is still initialised and `ForwardSDLLog` is still installed while the process unwinds and spdlog's static registry is destroyed. These are exactly the failure paths this change is meant to capture, so they matter more than the happy path.

Rather than adding a fourth and fifth call site, make it RAII (rule 8): a small scoped installer whose destructor restores SDL's default callback and drops the logger reference. Every exit path then gets it for free, and the two existing explicit calls go away.

### 2. `ForwardSDLLog` dereferences two globals unguarded

```cpp
g_defaultSDLLogOutput(nullptr, category, priority, message);
...
g_sdlLogger->log(...);
```

`RestoreSDLLogOutput()` calls `SDL_SetLogOutputFunction` and *then* `g_sdlLogger.reset()`. SDL logs from whichever thread produced the message, so a callback already in flight can observe the reset `shared_ptr` — and reading a `shared_ptr` while another thread writes it is a data race regardless. It is narrow (it follows `SDL_Quit()`), but an early `if (g_sdlLogger == nullptr) return;` and resetting before the `SDL_SetLogOutputFunction` call costs nothing. Same for `g_defaultSDLLogOutput` being null if the forwarder is ever installed twice or out of order.

### 3. `set_level(spdlog::level::trace)` diverges from every other logger

`MuLogger.cpp:38` creates every logger at `spdlog::level::info`. This one is forced to `trace`, so SDL debug/verbose output lands in `MuError.log` — which rotates at 5 MB × 3. With `SDL_LOGGING='video=verbose,gpu=verbose'` (which PR #602's harness sets by default) that can push the rest of the session's diagnostics out of the retained files, which defeats the purpose.

If the intent is "SDL's own category filter decides what reaches us, spdlog should not filter again", that is defensible — but rule 3 and rule 9 want it named and stated, not a bare literal. Otherwise use the same `info` default as the other loggers.

### 4. Rules 1, 6 and 7 — this belongs in its own file

84 lines of a self-contained subsystem (priority mapping, category naming, forwarder, install/restore) added to `Winmain.cpp`, which is already one of the larger monoliths in the tree. Rule 1 asks for extraction when you touch a big file; rule 7 spells out the shape for exactly this case — a new file with the `NewUI`-free bare name, free functions wrapped in `<Layer>::<Concern>`. Something like `src/source/Core/Utilities/Log/SdlLogBridge.{h,cpp}` under `namespace Core::Log::Sdl`, next to `MuLogger.{h,cpp}`, leaves `Winmain.cpp` with a single scoped-installer declaration. That also makes the RAII change in point 1 the natural shape rather than extra work.

### Smaller notes

- `SDL_LOG_CATEGORY_CUSTOM` and above hit the `default:` branch and print as `SDL category 19: …`. That is the documented fallback and fine, but the switch covers only SDL3's predefined set — a comment saying so would stop a future reader from thinking it is an omission.
- Rule 11: worth one line in `docs/build/linux/console.md` telling a reporter that SDL diagnostics now land in `MuError.log` and that `SDL_LOGGING` controls the verbosity. That is usage-level and is the whole point of the change.

Nothing here is structural — points 1 and 2 are small, 3 is a one-line decision, and 4 is a move. Happy to look again once they're in.

---

r0naldoom:
Addressed in ac53474a and force-pushed.

1. `Core::Log::Sdl::ScopedLogOutput` is constructed immediately after `mu::log::Init()` in `WinMain`. Its destructor drops the logger reference and restores the default SDL callback, so all later returns—including video, window, renderer, and font initialization failures—follow the same cleanup path. The former explicit restore calls are removed.
2. `SdlLogBridge.cpp` serializes the callback globals with a mutex, checks both pointers before use, and resets `g_sdlLogger` before `SDL_SetLogOutputFunction` restores the default callback. An in-flight forwarder therefore cannot race the shared-pointer reset.
3. `kSDLLogForwardingLevel` deliberately uses `trace`: `SDL_LOGGING` selects which diagnostics the user asks to retain, and the bridge must not filter those selected records a second time. The constant states that intent instead of hiding it in a literal.
4. Mapping, category naming, forwarding, and scoped lifetime now live in `src/source/Core/Utilities/Log/SdlLogBridge.{h,cpp}` under `Core::Log::Sdl`; `Winmain.cpp` only creates the scoped installer. The category switch now states that it covers SDL3's predefined categories and preserves the numeric fallback for custom values.

Also added the requested Linux console line: SDL diagnostics land in `MuError.log`, and `SDL_LOGGING` controls verbosity.

Verification:
- `cmake --preset linux-x64 && cmake --build --preset linux-x64-release && ctest --test-dir out/build/linux-x64 -C Release --output-on-failure` — 207/207 passed.
- Changed C++ files passed `clang-format --dry-run --Werror` and CI-argument `cppcheck`.
- New regression test `SDL bridge writes named diagnostics to MuError.log` installs the bridge, emits an SDL video log, and confirms the persisted `[sdl] [info] video:` record after teardown.

The scoped object owns only the SDL callback and its logger reference; `mu::log` lifetime is unchanged, `mu::log::Shutdown()` stays in `ShutdownRuntime()`.

r0naldoom:
Follow-up in ac53474a: `ScopedLogOutput` now scopes only the SDL callback and its logger reference; `mu::log` lifetime is unchanged, with `mu::log::Shutdown()` restored in `ShutdownRuntime()`. The regression test now calls `mu::log::Shutdown()` itself after the scoped bridge exits. Full Linux Release checks remain green: 207/207 tests passed.

sven-n:
## Re-review (`ac53474`)

All four points from the previous round are addressed the way I hoped: the bridge is its own file under `Core::Log::Sdl` with a `NewUI`-free name, `Winmain.cpp` is down to one declaration that covers every return, the forwarding level is a named constant with the reasoning next to it, both globals are null-checked, and there's a regression test plus the docs line. Good shape.

One new problem was introduced by the fix for point 2, and it needs to go before this merges.

### `g_sdlLogMutex` creates a lock-order inversion with SDL's own log lock

SDL holds `SDL_log_function_lock` *across the callback invocation*. From `src/SDL_log.c` at the pinned revision (`d9d5536`):

```c
// SDL_LogMessageV, line 642
SDL_LockMutex(SDL_log_function_lock);
{
    SDL_log_function(SDL_log_userdata, category, priority, message);
}
SDL_UnlockMutex(SDL_log_function_lock);
```

and `SDL_SetLogOutputFunction` takes the same mutex (line 833):

```c
void SDL_SetLogOutputFunction(SDL_LogOutputFunction callback, void *userdata)
{
    SDL_LockMutex(SDL_log_function_lock);
    { SDL_log_function = callback; SDL_log_userdata = userdata; }
    SDL_UnlockMutex(SDL_log_function_lock);
}
```

So the two acquisition orders are opposed:

| | first | then |
|---|---|---|
| any thread logging via SDL | `SDL_log_function_lock` | `g_sdlLogMutex` (in `ForwardSDLLog`) |
| thread running the ctor/dtor | `g_sdlLogMutex` | `SDL_log_function_lock` (in `SDL_SetLogOutputFunction`) |

That is a textbook deadlock. The practical window is narrow — the ctor runs before `SDL_InitSubSystem` and the dtor runs after `SDL_Quit()` on the main path — but it is wide open on the three early-return paths (`Winmain.cpp:1987`, `:2002`, `:2035`), where SDL video and the GPU backend are still up and can log from their own threads while `WinMain` unwinds. A hang on exit is a miserable bug to chase, and those failure paths are precisely what this change exists to capture.

The good news is the fix is a deletion. SDL's lock already gives you everything the mutex was added for:

- Only one callback invocation runs at a time.
- The callback cannot be swapped while it is running, so once `SDL_SetLogOutputFunction` returns, `ForwardSDLLog` is guaranteed not to be executing and not to be entered again.

So drop `g_sdlLogMutex` entirely and let the restore order carry the guarantee:

```cpp
ScopedLogOutput::ScopedLogOutput()
{
    g_defaultSDLLogOutput = SDL_GetDefaultLogOutputFunction();
    g_sdlLogger = mu::log::Get("sdl");
    g_sdlLogger->set_level(kSDLLogForwardingLevel);
    SDL_SetLogOutputFunction(ForwardSDLLog, nullptr);  // publishes the two globals
}

ScopedLogOutput::~ScopedLogOutput()
{
    SDL_SetLogOutputFunction(g_defaultSDLLogOutput, nullptr);  // no callback can be running once this returns
    g_sdlLogger.reset();
}
```

Note the destructor order flips: restore **first**, then release the logger. The current order (`reset()` then restore) means that during the gap SDL can still invoke `ForwardSDLLog`, which hits the `g_sdlLogger == nullptr` early return and silently drops those messages — the last few lines before shutdown, which are usually the interesting ones. Keeping the null checks as cheap defence is fine.

Worth a short comment on `ForwardSDLLog` recording *why* there is no lock, so nobody adds one back. Forwarding to `g_defaultSDLLogOutput` from inside the callback stays correct either way — `SDL_LogOutput` reads `SDL_priority_prefixes`, which `SDL_log.c:91` documents as guarded by the same `SDL_log_function_lock` the calling thread already holds.

### Minor, non-blocking

- The guard now outlives `mu::log`. `ShutdownRuntime()` calls `SDL_Quit()` then `mu::log::Shutdown()`, and `~ScopedLogOutput` only runs when `WinMain` returns one statement later (`Winmain.cpp:2231-2233`). Nothing logs in that gap, so it is harmless today, but the nesting is inverted from what the PR description says and from what the new test does — the test destroys the guard *before* `mu::log::Shutdown()`, which is the correct order. Worth making `WinMain` match, e.g. by scoping the guard so it dies before `ShutdownRuntime` runs, so the test and production agree.
- The three early-return paths never call `mu::log::Shutdown()` at all, so those logs are flushed only by spdlog's static teardown. Pre-existing, not yours — just noting it since it is the same set of paths.
- `SdlLogBridge.cpp` is picked up by the `GLOB_RECURSE` at `src/CMakeLists.txt:73`, so no build-file change is needed. Confirmed, not a request.

Drop the mutex, flip the destructor order, and I think this is done.

---

r0naldoom:
Addressed in eb87d797 and force-pushed.

- `g_sdlLogMutex` is gone. `ForwardSDLLog` now carries a comment stating why it takes no lock: SDL holds `SDL_log_function_lock` across the callback and across `SDL_SetLogOutputFunction()`, which already serializes invocations and guarantees none is running once the callback is restored. Both null checks stay as cheap defence.
- Destructor order flipped: restore the default callback first, then release the logger, so the last messages before teardown still reach the forwarder instead of hitting the early return.
- `WinMain` now matches the test's nesting. The guard is a `std::optional<ScopedLogOutput>` emplaced right after `mu::log::Init()`, and `ShutdownRuntime()` releases it after `SDL_Quit()` and before `mu::log::Shutdown()`. Early returns still restore it automatically through the destructor; the explicit release only orders the normal exit.

Verification on Linux x64 Release:
- clang-format 21.1.8 on the changed lines and cppcheck with the CI arguments: clean.
- `cmake --build --preset linux-x64-release` and `ctest --test-dir out/build/linux-x64 -C Release`: 207/207 passed, including `SDL bridge writes named diagnostics to MuError.log`.
- Not exercised: the three early-return paths with a live SDL logging thread; the guarantee there now rests on SDL's lock rather than on anything this PR adds.

sven-n:
## Re-review (`eb87d79`) — deadlock resolved, no blocking findings left

The lock-order inversion is gone and the replacement reasoning holds up. Walking through it rather than taking it on faith:

- **Publication of the globals is still ordered.** The constructor writes `g_defaultSDLLogOutput` and `g_sdlLogger` before calling `SDL_SetLogOutputFunction`, which acquires and releases `SDL_log_function_lock`. The logging thread acquires that same mutex before invoking the callback, so there is a release/acquire edge through SDL's own lock — the forwarder cannot observe half-initialised globals. Dropping the mutex did not trade the deadlock for a data race.
- **The destructor's new order is the one that matters.** `SDL_SetLogOutputFunction(g_defaultSDLLogOutput, nullptr)` returns only after it has taken the lock that `SDL_LogMessageV` holds across the invocation, so any in-flight `ForwardSDLLog` has completed and no further one can start. `g_sdlLogger.reset()` after that point is unobservable. The messages that previously fell into the `nullptr` early return now go through.
- **The comment on `ForwardSDLLog`** states the constraint rather than the conclusion, which is what stops someone reinstating the mutex. Good.

`std::optional<ScopedLogOutput>` also works out: the class is non-copyable and, because it declares a destructor, non-movable — `emplace()` constructs in place and `reset()` destroys in place, neither of which needs either operation, so the guard stays a true singleton owner. `ShutdownRuntime()` now releases it after `SDL_Quit()` and before `mu::log::Shutdown()`, so production and the regression test agree on the nesting, and the early returns still get the destructor for free. The added `<optional>` include is there.

Taking the second parameter on `ShutdownRuntime` purely to sequence the release is slightly awkward, but it is the honest way to express the ordering and I would not change it.

### Minor, non-blocking

- **The PR description is now stale.** It still says "A mutex protects the callback state from concurrent SDL log threads" and "destruction resets the logger reference before restoring SDL's callback" — both are the opposite of what `eb87d79` does. Worth fixing before merge so the squashed commit message doesn't describe code that isn't there.
- **`SDL_GetDefaultLogOutputFunction` vs. `SDL_GetLogOutputFunction`.** The bridge captures SDL's *built-in* default and restores that, rather than whatever callback was installed beforehand. Nothing else in the tree installs one, so the behaviour is identical today — but now that this is a reusable `Core::` component, `SDL_GetLogOutputFunction(&previous, &previousUserdata)` would preserve a prior callback and its userdata instead of silently replacing it with the default. It would also make the forwarder's `g_defaultSDLLogOutput(nullptr, …)` correct for a callback that actually reads its userdata; right now that `nullptr` is only safe because SDL's default ignores it.
- **Nesting would silently disarm the bridge.** Two live `ScopedLogOutput` instances would share the same file-scope globals: the inner destructor restores the default and nulls the logger while the outer guard is still alive, leaving the bridge off with no indication. Nobody does this today, and the type is clearly intended as a single scoped owner — an `SDL_assert` or early return on `g_sdlLogger != nullptr` in the constructor would make that intent enforceable rather than conventional.

Nothing here blocks. Fix the description and this is good to go — say the word and I'll merge it. Same caveat as #601 and #602: no CI has run on this branch (0 check runs), and unlike those two this one changes C++ that builds on all three platforms, so it's the one I'd most want a green Windows and macOS build on before it lands.

---

r0naldoom:
Description rewritten to match the code (no mutex, restore before release, `std::optional` guard released in `ShutdownRuntime()`).

I also took the two non-blocking points, since they are cheap and make the `Core::` component correct on its own terms (f5c514c7):

- The bridge now captures the previously installed callback and its userdata with `SDL_GetLogOutputFunction`, forwards to that pair, and restores that pair on destruction, instead of the built-in default with a `nullptr` userdata.
- A nested `ScopedLogOutput` stays inert (`m_installed` is only set by the instance that installed the bridge), so it no longer disarms the outer guard. The regression test now creates and destroys a nested guard between two SDL messages and checks that both reach `MuError.log`.

Gates on Linux x64 Release: clang-format on the changed lines, cppcheck with the CI arguments, build, 207/207 tests.

On CI: the fork has not managed to run the workflow yet, so there is still no Windows or macOS build from my side. If the checks on this PR are waiting on an approval for fork workflows, approving them here would be the quickest way to get those two builds before merge. Otherwise, ready to merge from my side.

PR #601, opened 2026-09-20
build(linux): add opt-in static SDL backend and Vulkan loader linkage

r0naldoom:
## Summary

Adds opt-in `MU_LINK_SDL_PLATFORM_BACKENDS` for Linux packages whose runtime does not expose SDL's dynamically loaded video backends or the Vulkan loader through a global library path.

## Root cause

SDL's default Linux video backends dynamically load their system libraries. The client also loads Vulkan at runtime. A packaged executable can therefore start without those libraries in its runtime search path even when the build environment resolved them.

## Fix

- `src/CMakeLists.txt:22` adds the Linux-only, default-OFF option and forces SDL's X11, Wayland, libdecor, and KMSDRM backends to link their dependencies statically when enabled.
- `src/CMakeLists.txt:250` links `Vulkan::Vulkan` to make the loader a runtime dependency with its CMake-provided search path.
- `docs/build/README.md:75` and `docs/build/linux/console.md:57` document the option.

## Related issues

None open upstream. Companion to #599 (which now names the driver and SDL error when this failure happens).

## Test plan

- [x] Fresh default-OFF Linux configure and Release build; 206 CTest cases passed, and `ldd Main` did not list `libX11.so.6`, `libwayland-client.so.0`, `libdecor-0.so.0`, or `libvulkan.so.1`.
- [x] Fresh `-DMU_LINK_SDL_PLATFORM_BACKENDS=ON` Linux configure and Release build; `ldd Main` resolved `libX11.so.6`, `libwayland-client.so.0`, `libdecor-0.so.0`, and `libvulkan.so.1`.
- [x] Launched the enabled build without `LD_LIBRARY_PATH`; startup log recorded `Start window success`, Vulkan device selection, and GPU initialization.
- [x] Project pre-PR verification completed: configuration, build, and 206 CTest cases passed.
- [ ] Windows and macOS were not built; the option has no effect outside Linux through its `CMAKE_SYSTEM_NAME` guard.

- [x] Toggle in one build directory: configure with `-DMU_LINK_SDL_PLATFORM_BACKENDS=ON` (cache: the four `SDL_*_SHARED` = `OFF`), reconfigure the same directory with `=OFF` (cache: all four back to `ON`), build: `ldd Main` shows none of `libX11`/`libwayland`/`libdecor`/`libvulkan`; reconfigure `=ON`, build: `ldd Main` resolves `libX11.so.6`, `libwayland-client.so.0`, `libdecor-0.so.0`, `libvulkan.so.1`; 206/206 tests pass.

## Checklist

- [x] The default build behavior remains unchanged.
- [x] Documentation covers the opt-in Linux configuration.
- [x] Change is limited to Linux build linkage and its documentation.


sven-n:
## Code review

Reviewed against [`docs/CODING_RULES.md`](../blob/main/docs/CODING_RULES.md). The approach is sound — opt-in, Linux-guarded, default-OFF, documented in both the option table and the console guide. Three things to fix.

### 1. The option is not reversible in an existing build directory

```cmake
if (CMAKE_SYSTEM_NAME STREQUAL "Linux" AND MU_LINK_SDL_PLATFORM_BACKENDS)
  set(SDL_X11_SHARED OFF CACHE BOOL "" FORCE)
  ...
endif()
```

`FORCE` writes the cache on every configure while the option is ON. When someone then reconfigures the *same* build directory with `-DMU_LINK_SDL_PLATFORM_BACKENDS=OFF`, the branch simply doesn't run — nothing restores the four `SDL_*_SHARED` variables, so they stay `OFF` in the cache and SDL keeps linking its backends directly. The checklist item "the default build behavior remains unchanged" holds for a fresh configure but not for a toggled one, and the failure is silent.

Add the `else()` branch that forces them back `ON`, so the option is symmetric:

```cmake
if (CMAKE_SYSTEM_NAME STREQUAL "Linux" AND MU_LINK_SDL_PLATFORM_BACKENDS)
  ...
else()
  set(SDL_X11_SHARED ON CACHE BOOL "" FORCE)
  ...
endif()
```

(Or state in the docs that changing the option requires a fresh build directory — but the `else()` is cheaper than a footgun.)

### 2. `Fixes #4` is the wrong issue

[#4](https://github.com/sven-n/MuMain/issues/4) is a closed 2023 PR by @sven-n about dynamic frame rate — unrelated to SDL backend linkage. If you meant an issue in your fork, link it by full URL; otherwise drop the line so the merge doesn't imply an issue was resolved.

### 3. Docs wording overstates what happens

> links SDL video backends and the Vulkan loader into `Main`

Nothing is linked *into* `Main`. `SDL_*_SHARED=OFF` makes SDL link against `libX11.so.6`, `libwayland-client.so.0`, `libdecor-0.so.0` and `libdrm`/`libgbm` at build time instead of `dlopen`-ing them at run time — they become `DT_NEEDED` entries with a recorded `RUNPATH`, which is precisely what your `ldd` results show. Suggested phrasing for the option table:

> Linux only: makes SDL link its video-backend libraries and the Vulkan loader at build time instead of loading them at run time, so a packaged runtime resolves them without a global library path. Defaults to `OFF`.

### Smaller notes

- `find_package(Vulkan REQUIRED)` sits at the `target_link_libraries(Main …)` site rather than with the other dependency discovery near the top. Works, but a reader scanning for the project's dependencies won't find it. Not blocking.
- Worth confirming that `Vulkan::Vulkan` records a `RUNPATH` that survives packaging — on most distros `libvulkan.so.1` lives in the default loader path, so the `find_package` may contribute nothing beyond the `DT_NEEDED` entry. That is still the useful half of the change; just make sure the docs don't promise more than the link line delivers.

Point 1 is the one that matters; 2 and 3 are text.

---

r0naldoom:
Thanks for the review. All three addressed in the amended commit (`0de8be5d`):

1. **Reversible toggle**: the block now has the `else()` branch forcing the four `SDL_*_SHARED` back `ON`, with a comment saying why both branches `FORCE` the cache. Verified in one build directory: configure `=ON` (cache: all four `OFF`), reconfigure `=OFF` (cache: all four `ON`), build, `ldd Main` shows none of `libX11`/`libwayland`/`libdecor`/`libvulkan`; reconfigure `=ON`, build, `ldd Main` resolves `libX11.so.6`, `libwayland-client.so.0`, `libdecor-0.so.0`, `libvulkan.so.1`. 206/206 tests pass in both states.
2. **`Fixes #4`** removed from the body; it referred to an issue in my fork, not to this repository.
3. **Wording**: option description, CMake comments, the option table and the console guide now say what happens: SDL links its video-backend libraries and the Vulkan loader at build time instead of `dlopen`-ing them, so they become `DT_NEEDED` entries with a recorded `RUNPATH`. Nothing is described as linked into `Main`.

Smaller notes: `find_package(Vulkan)` moved next to the other dependency discovery (`OpenSSL`, `CURL`), with the `target_link_libraries` staying at the `Main` target. On the `RUNPATH`: agreed that on most distributions `libvulkan.so.1` is on the default loader path and the `DT_NEEDED` entry is the useful half; the docs only promise resolution through the recorded `RUNPATH` where one exists.

sven-n:
## Re-review (`0de8be5`) — all points addressed, merging

- **Reversibility.** The `else()` branch is there and both sides `FORCE`, so toggling in an existing build directory is symmetric now. I checked the four knobs against SDL at the pinned revision (`d9d5536`, `CMakeLists.txt:341`, `:356`, `:358`, `:376`) — `SDL_X11_SHARED`, `SDL_WAYLAND_SHARED`, `SDL_WAYLAND_LIBDECOR_SHARED` and `SDL_KMSDRM_SHARED` all default to `ON`, so the `else()` branch writes exactly SDL's own defaults and the default build is genuinely unchanged. They are `dep_option`s, so SDL still forces them `OFF` when the owning backend is disabled; pre-setting `ON` doesn't interfere with that. Your toggle test matches.
- **`Fixes #4`.** Gone, replaced with the #599 cross-reference, which is the accurate relationship.
- **Docs.** Both the option table and the console guide now describe what actually happens — build-time linkage with a recorded `RUNPATH` rather than "linked into `Main`" — and the console guide ties it to the concrete `SDL video init failed` / `wayland not available` symptom, which is the thing a packager will search for. The comment in `src/CMakeLists.txt` explaining the `dlopen`-vs-`DT_NEEDED` tradeoff is a good addition.
- `find_package(Vulkan REQUIRED)` moved up next to the `OpenSSL` discovery with `target_link_libraries(Main …)` left at the link site. Same directory scope, so the imported target is visible where it's used.

One non-blocking note: on Linux the `else()` branch now `FORCE`s those four variables on every configure, so a user passing `-DSDL_X11_SHARED=OFF` by hand will find it silently overridden. Given SDL is vendored and this option is the supported way to ask for that, taking ownership of the knobs is the right call — just be aware it's now the project's decision rather than SDL's.

---

PR #602, opened 2026-09-20
feat: add Wayland diagnostics capture harness

r0naldoom:
## Summary

Adds `scripts/wayland-diag.sh`, a dependency-light capture harness for one Linux client invocation. It preserves raw evidence locally and writes a sanitized, issue-ready `summary.txt`.

## Root cause

Current Linux graphics reports depend on a user's prose description and ephemeral standard error. Maintainers cannot reliably distinguish compositor protocol support, SDL backend fallback, Vulkan setup, and process outcome from that evidence.

## Fix

- `scripts/wayland-diag.sh:41` provides documented timeout, client selection, output, archive, and headless options.
- `scripts/wayland-diag.sh:183` records compositor state, selected SDL/Vulkan/Wayland/Mesa variables, tool availability, repository and SDL revisions, optional Wayland globals, and Vulkan summary without installing tools.
- `scripts/wayland-diag.sh:412` captures `WAYLAND_DEBUG=client`, non-protocol standard error, Vulkan validation messages, `MuError.log`, and an explicit process classification.
- `scripts/wayland-diag.sh:495` runs the headless Weston/lavapipe path only after preflight succeeds, then tears Weston down through the exit trap.
- `scripts/wayland-diag.sh:610` normalizes an Environment / Steps / Result summary without timestamps, absolute paths, or selected sensitive values.
- `docs/build/README.md:76`, `docs/build/linux/console.md:78`, and `.gitignore:30` document reporting and keep local bundles untracked.

## Test plan

- [x] Adjusted Linux build with `SDL_VIDEO_DRIVER=wayland` and a 30-second capture: `summary.txt` reported fifo-v1 state, Wayland protocol traffic, Vulkan driver/device, and `timed-out`.
- [x] Default Linux build with a 30-second capture: `exit-status.txt` recorded `exited 0`; `summary.txt` preserved `SDL video init failed` as the result instead of treating exit code zero as successful graphics initialization.
- [x] Headless Weston session with lavapipe: the summary recorded Wayland protocol traffic, Vulkan, llvmpipe, and `timed-out`.
- [x] Headless run without Weston on `PATH`: preflight returned `failed-to-start` before the client and listed `weston` as missing.
- [x] Timed mock capture: completed with status 0 while `exit-status.txt` retained `classification: timed-out`; preserved a prior `MuError.log` as `MuError.log.previous`; routed both supported libwayland record formats.
- [x] Wayland client capture: `summary.txt` reported `configure=3`, `ack_configure=1`, `commits=3`, and `frame callbacks=1`.
- [x] `shellcheck scripts/wayland-diag.sh` completed without findings.

## Checklist

- [x] Raw logs and optional archive remain local; public reporting starts from the sanitized summary.
- [x] Normal mode preserves user SDL/Vulkan/Wayland/Mesa settings; headless mode supplies its private display and lavapipe ICD.
- [x] No C++ code or CI scope changed.


sven-n:
## Code review

Reviewed against [`docs/CODING_RULES.md`](../blob/main/docs/CODING_RULES.md). The script is well-structured — short single-purpose functions, `set -euo pipefail`, an `EXIT` trap that tears Weston down, mode `100755`, and documentation in both `docs/build/README.md` and the console guide. The problems are in the capture and classification logic, where a diagnostics tool has to be right or it is worse than nothing.

### 1. The summary can be built from a half-written log (race)

```bash
) 2> >(split_client_stderr) &
client_pid=$!
```

Bash does not wait for a process substitution. `run_client` returns as soon as `wait "$client_pid"` returns, and `main` goes straight into `write_summary`, which greps `wayland-debug.log`, `stderr.log` and `vulkan-validation.log`. `split_client_stderr` is very likely still draining at that moment — with `WAYLAND_DEBUG=client` it is processing thousands of lines through a `while read` loop with three `printf`s per line, which is slow. So the "registry binds", protocol counts and first-warning/first-error fields are computed from whatever happened to be flushed.

This is not theoretical; it will bite on any run with real protocol traffic, and it fails *quietly* — the summary looks complete. Simplest fix: send stderr to a raw file, `wait` for the client, then split the file afterwards. If you want to keep streaming, use a named FIFO with a reader whose PID you capture and `wait` on before summarising.

### 2. Three of the four protocol counters are always zero

```bash
"$(count_protocol_calls '\.configure\()')" \
"$(count_protocol_calls '\.ack_configure\()')" \
"$(count_protocol_calls '\.commit\()')" \
"$(count_protocol_calls '\.frame\()')"
```

The pattern matches a literal *empty* argument list. `WAYLAND_DEBUG` prints arguments, so the real lines are:

- `xdg_surface@N.configure(12345)` — has a serial → never matches
- `xdg_surface@N.ack_configure(12345)` — has a serial → never matches
- `wl_surface@N.frame(new id wl_callback@M)` — has an argument → never matches
- `wl_surface@N.commit()` — no arguments → the only one that works

So every report will say `configure=0; ack_configure=0; commits=<n>; frame callbacks=0`, and a maintainer reading that would reasonably conclude the client never got configured — the opposite of the truth. Drop the closing paren from all four: `'\.configure\('`. (The stray unescaped `)` is also not valid ERE; GNU grep tolerates it, other greps need not.)

### 3. Protocol-line detection only matches recent libwayland

```bash
[[ "$1" =~ ^\[[0-9]{2}:[0-9]{2}:[0-9]+\.[0-9]+\][[:space:]]+\{ ]]
```

This requires both the wall-clock `[HH:MM:SS.mmm]` stamp *and* the `{queue}` segment. Older libwayland prints `[ 1234567.890] wl_display@1.get_registry(new id wl_registry@2)` — elapsed microseconds, no queue name. On those systems every protocol line falls through to `stderr.log`, `wayland-debug.log` stays empty, and the summary reports `video backend evidence: unknown`, `registry binds [none observed]` and zeros across the board.

That is the failure mode that hurts most, because the distros still shipping the older format are exactly the ones a struggling user is likely to be on. Make the timestamp accept both shapes and the `{queue}` segment optional.

### 4. `rm -f "$muerror_source"` destroys the user's existing log

Before the run, the script deletes `MuError.log` next to the client. A user whose client just crashed, who then reaches for the diagnostics harness as instructed by `docs/build/linux/console.md`, loses the log from the crash they are trying to report. It is mentioned nowhere in `--help` or the docs.

Copy it into the bundle first (`MuError.log.previous`), or rename it aside rather than unlinking it. Starting from a clean file for the run is the right call — silently discarding the old one is not.

### 5. A normal successful capture exits non-zero

```bash
if [[ "$process_classification" == timed-out ]]; then
    return 124
fi
```

With a fixed `--timeout`, a client that starts and runs correctly *always* ends as `timed-out` — that is the expected outcome, not a failure. Any wrapper with `set -e`, or a user checking `$?`, will read a good capture as a failed one. Consider reserving non-zero for "the capture could not be made" (`failed-to-start`, preflight failure) and letting a completed capture exit 0 regardless of how the client was stopped; the classification is already in `summary.txt` and `exit-status.txt` where it belongs.

### Smaller notes

- `VK_INSTANCE_LAYERS` is deprecated in current Vulkan loaders in favour of `VK_LOADER_LAYERS_ENABLE`. Setting both would keep validation working across loader versions.
- No `shellcheck` job exists in `.github/workflows/ci.yml`, so you verified it locally but nothing keeps it clean. Adding it to the existing lint job is a few lines and would cover `scripts/` generally, not just this file — reasonable as a separate PR.
- `capture-method.txt` documenting the routing rule is a nice touch; once #3 is fixed it would be worth noting there which libwayland formats are recognised.
- `first_line_matching` always returns 0 and signals "not found" via empty output. That is fine and used consistently, but it makes `|| true` at the call sites unnecessary noise.

Points 1–4 all cause a confidently-worded but wrong `summary.txt`, which is the one thing this harness cannot afford. Worth another look once they're addressed.

---

r0naldoom:
Addressed in dd3e2b69 and force-pushed.

1. Client stderr now goes to `client-stderr.raw.log`; `run_client` waits for the client and only then calls `split_client_stderr`, before `write_summary`. The summary cannot observe a half-written routed log.
2. Counter patterns now match the opening parenthesis. A Wayland client capture reports: `configure=3; ack_configure=1; commits=3; frame callbacks=1`.
3. Routing accepts both `[HH:MM:SS.mmm] {queue}` and `[ 1234567.890]` records. `capture-method.txt` documents both forms.
4. A prior client-directory `MuError.log` is copied as `MuError.log.previous` before the fresh capture; the Linux reporting guide now says so.
5. A completed capture returns 0 even when its recorded classification is `timed-out`. The timed mock run returned 0 and its `exit-status.txt` contained `classification: timed-out` with raw exit code 143.

Smaller points:
- When validation is enabled, the client receives both `VK_INSTANCE_LAYERS=VK_LAYER_KHRONOS_validation` and `VK_LOADER_LAYERS_ENABLE=VK_LAYER_KHRONOS_validation`; either user-provided variable is preserved.
- Summary call sites use the successful empty result from `first_line_matching` directly. Remaining `|| true` guards handle intentional non-zero external probes under `set -e`.
- `shellcheck scripts/wayland-diag.sh` is clean. CI scope remains unchanged because repository-wide shell linting is a separate concern.

Verification:
- `shellcheck scripts/wayland-diag.sh`
- `cmake --preset linux-x64 && cmake --build --preset linux-x64-release && ctest --test-dir out/build/linux-x64 --output-on-failure` — 206/206 passed.
- Wayland capture excerpt above was produced from the supplied runtime client.

r0naldoom:
Addendum with a real-client run of the amended script (`dd3e2b69`), Linux x64 Release with the SDL backends linked, COSMIC, automatic video driver, `--timeout 30`:

- Routing: 3635 protocol records in `wayland-debug.log`, 6 SDL lines in `stderr.log` (the fifo-v1 fallback, the failed X11 connection, then `SDL_GPU Driver: Vulkan` / `Vulkan Device: NVIDIA GeForce GTX 1650`), nothing interleaved.
- Counters, `summary.txt` vs `grep -Ec` on the raw log: `configure=7`, `ack_configure=3`, `commits=190`, `frame callbacks=317`, identical.
- A stale `MuError.log` placed beside the client before the run came out as `MuError.log.previous` in the bundle; the fresh `MuError.log` holds this run.
- The client exited on its own this time; `exit-status.txt`: `classification: exited 0`, script exit code 0. A separate timed run classified `timed-out` also returned 0; the headless preflight without Weston still returns 2.

sven-n:
## Re-review (`dd3e2b6`) — all points addressed, merging

Every one of the five findings is fixed, and fixed properly rather than papered over:

- **Capture race.** The client's stderr now goes straight to `client-stderr.raw.log`, and `split_client_stderr` reads that file after `wait` returns. No process substitution, nothing to race. Keeping the raw stream as its own artifact is a bonus — the routing is now reproducible after the fact instead of being a one-shot decision.
- **Protocol counters.** `'\.configure\('`, `'\.ack_configure\('`, `'\.commit\('`, `'\.frame\('` — the trailing paren is gone, so lines with arguments match, and the patterns are valid ERE now. Your `configure=3, ack_configure=1, commits=3, frame callbacks=1` test result is the confirmation I wanted.
- **libwayland format coverage.** `is_wayland_protocol_line` takes both the `[HH:MM:SS.mmm] {queue}` prefix and the older `[ 1234567.890]` one, with the queue segment optional. That is the case that mattered most, since the distros still on the older libwayland are the ones most likely to be filing these reports.
- **`MuError.log`.** `preserve_existing_muerror` copies to `MuError.log.previous` before removing, and `docs/build/linux/console.md` tells the reporter it happens. The crash log someone came to report survives now.
- **Exit status.** `timed-out` returns 0; only `failed-to-start` returns 2. A successful capture of a healthy client no longer looks like a failure to `set -e` wrappers.

`VK_LOADER_LAYERS_ENABLE` alongside `VK_INSTANCE_LAYERS` (with `configure_validation` checking both before deciding it's a user setting) covers the loader-version spread, and `capture-method.txt` documents the new two-format routing so the bundle explains itself.

One cosmetic nit for whenever you next touch the file: `wayland_debug_file` is assigned in `prepare_output_directory` but, unlike its siblings, is missing from the declaration block at the top. Every read happens after `prepare_output_directory`, so `set -u` never fires — it just breaks the file's own convention and is a latent trap if a future caller reads it earlier.

The `shellcheck`-in-CI suggestion from last round still stands as a separate change if you want it; nothing here depends on it.

---

PR #606, opened 2026-09-20
fix(platform): close the game window on the first X / Alt+F4 press

### LINKED ISSUE #535: Game window is not closing on presing X
The game does not close at the first press of the X button Close button on the top right of the window when in Window mode.

When pressing again it closes

Expected would be to be closed at the first click.

Mosch0512:
## Summary

Fixes #535 - the game window in windowed mode needed two clicks on the
titlebar X to close. Two separate issues combined to cause it:

- `MainLoop()` only reacted to `SDL_EVENT_QUIT`, relying on SDL's implicit
  escalation of a window close request into that event. It now also handles
  `SDL_EVENT_WINDOW_CLOSE_REQUESTED` directly, so the close is acted on
  immediately on the first press.
- Once that was fixed, shutdown still looked hung: `ShutdownRuntime()` hid
  the window before `mu::WaitForSDLGpuIdle()` finished waiting on the last
  frame's present fence. A hidden/occluded window's swapchain isn't
  serviced by the compositor, so the wait could stall until unrelated
  desktop activity (e.g. a focus change) forced a compositor redraw.
  Waiting for the GPU to go idle before hiding the window lets the last
  frame present normally first, avoiding the stall.

## Test plan

- [x] Built `Main` (Release, mueditor preset) - compiles cleanly.
- [x] Launched the built executable and posted a single close request
      (equivalent to one titlebar-X click) twice independently; the process
      exited within ~0.5-1.4s each time with no extra interaction needed.
- [x] Confirmed (before this fix) that the window stayed visible/hung and
      required a second click or unrelated desktop interaction (e.g.
      switching focus) to actually close.


alandarev:
indeed, it must close fast before mum walks in and notices you still grinding elite yeti at 5AM before school.
Sorry for this joke, but it unlocked my childhood memories

PR #609, opened 2026-09-22
fix(render): keep the Lost Tower fire layer inside the strip cutouts

### LINKED ISSUE #589: Lost Tower walls render red/orange instead of dark stone
## Description

In Lost Tower, wall geometry renders with an incorrect bright red/orange tint in current builds, where it previously rendered as dark stone with only a thin red glow strip. Screenshot comparison shows the difference on the walls.

## Expected behavior

Walls render dark (stone/wood diffuse with baked dungeon darkening), matching the older build: dark blue-gray blocks with a red glow line.

## Actual behavior

Walls render bright red/orange with mottled detail across all visible wall segments. Floor, characters, and skill effects look normal.

## Environment

- 6.18.49-1-MANJARO (linux618)
- Build: current `main` (post-`6dda3643` merge; SDL GPU + HLSL pipeline)
- Map: Lost Tower (World/Object 8)

## Current
Image

## Expected
Image

r0naldoom:
## Summary

Lost Tower wall blocks render dark stone with the thin red glow lines again, instead of the whole block face covered in orange. Fixes #589.

| `main` (RenderDoc render target of the captured frame) | this PR |
|---|---|
| Lost Tower walls on main: block faces covered in orange | Lost Tower walls with this PR: dark stone, thin red lines on the strip |

## Root cause

For Lost Tower object types 19/20 and 3/4, `Draw_RenderObject` draws each block twice: the stone body, then `RenderBody(RENDER_TEXTURE | RENDER_BRIGHT, ..., BITMAP_CHROME)` with `BodyLight = (1.0, 0.2, 0.1)` and the texture streaming on one mesh. On the SDL GPU renderer that second pass is what the flags ask for: additive, drawn after the stone, on every non-hidden mesh, so the whole face receives the chrome wave texture tinted orange.

The look players know comes from the previous OpenGL renderer, which drew that pass with blending disabled and before the stone body. The stone then covered the fire, and the glow only survived where the strip texture (`re02.tga`, binary alpha) is discarded, which is the thin red line. RenderDoc captures of the same wall on both renderers confirm it: on the old build the render target is fully red after the fire draws and dark after the stone draws; on `main` a wall pixel goes from `(0.03, 0.07, 0.06)` after the stone to `(0.99, 0.46, 0.25)` after the additive fire draw. Texture decode, upload and binding were checked and are correct; the original client's `RenderMesh` uses the same additive math, so this is the previous renderer's layering that the port lost, not a texture or lighting error.

## Fix

- `RenderLostTowerFireBlock()` draws the stone body first and the fire layer after it as an opaque textured pass. With the renderer's depth test the fire is rejected on the faces the stone already covered and kept where the strip texture was discarded, which reproduces the previous layering without depending on draw order. The two duplicated blocks (types 19/20 and 3/4) become one helper; the fire tint is a named constant; the dead `Light`/`p`/`Luminosity` locals in that branch are gone.
- Two small commits make `ZzzObject.cpp` pass the CI static-analysis gate, which any change to this file would otherwise fail: a `VectorCopy` macro statement was missing its semicolon (cppcheck aborts the file on it), and the two `.obj` record readers used unaligned pointer casts (`*(short*)(Data + ptr)`) that cppcheck reports as warnings once it can parse the file; they now read with `memcpy` like the neighbouring fields.

Other models that use `RENDER_BRIGHT + BITMAP_CHROME` over the whole body (Blood Bone, Devil) are untouched.

## Test plan

- [x] Linux x64 Release: `cmake --preset linux-x64 && cmake --build --preset linux-x64-release && ctest --test-dir out/build/linux-x64 -C Release --output-on-failure`: clean build, 206/206 tests pass.
- [x] `clang-format --dry-run --Werror` (21.1.8) on the changed lines and `cppcheck` with the CI arguments on `ZzzObject.cpp`: clean (the file did not pass cppcheck before the two chore commits).
- [x] Lost Tower in-game on Linux/Vulkan: walls show dark stone with thin red lines on the strip, no orange on the faces; Blood Bone still has its red glow. Before/after above. The visual check was done before the fire tint literal was moved into the named constant; the values are identical.
- [x] Map loading after the reader change: the final build (all three commits) starts in Lorencia and warps into Lost Tower with the objects in place and the fixed walls.
- [ ] Windows and macOS not run; the change is renderer-independent (draw order and flags only).

## Checklist

- [x] Changes are focused on one concern per commit; the diff is as small as it reasonably can be.


r0naldoom:
Post-merge measurement, in case it is useful as a record.

I compared a capture of this build against a capture of the OpenGL client at `a410f3de`, on the same Lost Tower scene: fixed 800x600 window, fixed camera pose, fixed world time, pinned static-object luminosity, and the same frame anchor on both sides (a debug label emitted right after the static objects finish, so the compared render target precedes HUD drawing).

| Region | Before this change | After |
| --- | ---: | ---: |
| Wall ROI | 34.85% differing pixels | 0.0053% (4 px of 75,600) |
| Full anchored frame | 21.64% | 0.05% |

In the wall region the mean per-channel delta is 0.26 and p95 is 1; the four remaining pixels are isolated single-pixel components. The two runs differ only in this change -- scene, camera, resolution, data and instrumentation are identical, and the client under test also carries a local world-viewport adjustment, unrelated to this PR, present in both runs.

So the additive pass was the whole of it: with the fire layer drawn opaque and left to the depth test, the block matches the historical renderer to four pixels.

